Issue #11, opened 2020-03-04
Less than operator in LEFT JOIN


jdposada:
Hi,

In the query below a `<` operator is used in the JOIN. This is not supported in some databases, particularly in BQ. This query is in `CohortDiagnostics/inst/sql/sql_server/OrphanCodes.sql" `

```sql
SELECT ss1.*
INTO #search_string_subset
FROM #search_str_top1000 ss1
LEFT JOIN #search_str_top1000 ss2
	ON ss2.concept_name_length < ss1.concept_name_length
		AND ss1.concept_name LIKE CONCAT (
			'%',
			ss2.concept_name,
			'%'
			)
WHERE ss2.concept_name IS NULL;
```

I think that changing to this should be equivalent and work

```sql
SELECT ss1.*
INTO #search_string_subset
FROM #search_str_top1000 ss1
LEFT JOIN #search_str_top1000 ss2
	ON ss2.concept_name_length = ss1.concept_name_length
		AND ss1.concept_name LIKE CONCAT (
			'%',
			ss2.concept_name,
			'%'
			)
WHERE ss2.concept_name IS NULL
AND ss2.concept_name_length < ss1.concept_name_length;
```

schuemie:
Your proposal completely changes the logic. The goal of the query is to find search strings for which there are no shorter search strings that are a proper substring of the first search string. Your modification finds an empty set (I think), because the left join will always find the string itself and is therefore not NULL.

Maybe use a nested inner join inside the left join?

jdposada:
Thanks for the quick answer :)

An example:

thoracic spondylarthritis | spondylarthritis
postmeningococcal arthritis (disorder) | postmeningococcal arthritis

Using this query
```sql
SELECT
  *
FROM
  `som-nero-phi-rit.EhdenRaDrugUtilization.xyvfoksdsearch_str_top1000`
WHERE
  concept_name NOT IN (
  SELECT
    ss1.concept_name
  FROM
    `som-nero-phi-rit.EhdenRaDrugUtilization.xyvfoksdsearch_str_top1000` ss1
  CROSS JOIN
    `som-nero-phi-rit.EhdenRaDrugUtilization.xyvfoksdsearch_str_top1000` ss2
  WHERE
    ss1.concept_name LIKE CONCAT ( '%', ss2.concept_name, '%' )
    AND ss2.concept_name_length < ss1.concept_name_length)
```

[...]

PR #12, opened 2020-03-04
fixing a less than in a left join

jdposada:


jdposada:
Linked to this reported issue. 

https://github.com/OHDSI/CohortDiagnostics/issues/11

I tested in my BQ DB and it worked

PR #29, opened 2020-04-02
Replaces left join with not in

### LINKED ISSUE #11: Less than operator in LEFT JOIN

wivern:
Fixes #11

### DIFF
--- a/inst/sql/sql_server/OrphanCodes.sql
+++ b/inst/sql/sql_server/OrphanCodes.sql
@@ -129,14 +129,13 @@ WHERE rn1 < 1000;
 SELECT ss1.*
 INTO #search_string_subset
 FROM #search_str_top1000 ss1
-LEFT JOIN #search_str_top1000 ss2
-	ON ss2.concept_name_length < ss1.concept_name_length
-		AND ss1.concept_name LIKE CONCAT (
-			'%',
-			ss2.concept_name,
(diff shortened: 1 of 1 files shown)


schuemie:
Tested on RedShift, and seems to work.

Issue #35, opened 2020-04-07
error when generateInclusionStats = FALSE but SQL has inclusionStats


gowthamrao:
This issue is when we are instantiating and generating cohort statistics based on JSON/SQL specifications in package - created using `ROhdsiWebApi::insertCohortDefinitionInPackage` function. 

If the specifications were created using `ROhdsiWebApi::insertCohortDefinitionInPackage1 function with `generateStats = TRUE`, sql will have snippets on inclusion stats. These snippets use 'results_database_schema', not 'target_database_schema'.

If these specifications are called by function generateInclusionStats of CohortDiagnostics with generateInclusionStats = FALSE - it will throw an error because SqlRender is missing results_database_schema.

> 
> Error:
> java.sql.SQLException:  Invalid operation: syntax error at or near "@" 
> Position: 13;
> 
> SQL:
> delete from @results_database_schema.cohort_inclusion_result where cohort_definition_id = 15234 and mode_id = 0

Error from Function `instantiateCohortSet' https://github.com/OHDSI/CohortDiagnostics/blob/29aaaf756e86f1a285461c7cba220cb2f5894675/R/CohortConstruction.R#L429 

https://github.com/OHDSI/CohortDiagnostics/blob/29aaaf756e86f1a285461c7cba220cb2f5894675/R/CohortConstruction.R#L523

schuemie:
Yes, this is unfortunately unavoidable: if you imported the cohort definitions with the inclusion rule statistics logic embedded in the SQL, you must run the inclusion stats when instantiating them.

It would be good to mark this clearly in the documentation, and throw a meaningful error when the user tries to do it differently anyway. The best way to detect whether the SQL includes the inclusion rule stats is probably to do a regex search for one of the statistics tables that is required. If the table is there, but the user has inclusion statistics turned of, we should throw a clear error.

@gowthamrao : would like to give that a try?

gowthamrao:
yes. i will assign this to myself, but will combine this as part of a broader release.

gowthamrao:
if `runInclusionStatistics = TRUE`, we either need a default (not NULL) value for `inclusionStatisticsFolder` or we need user to provide it.

PR #45, opened 2020-04-19
Throws a meaningful warning if there is a mismatch between sql and options for generate inclusion statistics

gowthamrao:
Throws a meaningful warning if there is a mismatch between sql and options for generate inclusion statistics

Fixed https://github.com/OHDSI/CohortDiagnostics/issues/35

### DIFF
--- a/R/CohortConstruction.R
+++ b/R/CohortConstruction.R
@@ -180,7 +180,7 @@ instantiateCohort <- function(connectionDetails = NULL,
   
   ParallelLogger::logInfo("Instantiation cohort with cohort_definition_id = ", cohortId)
   sql <- cohortSql
-  .warnMismatchSqlInclusionStats(sql)
+  .warnMismatchSqlInclusionStats(sql, generateInclusionStats = generateInclusionStats)
   if (generateInclusionStats) {
     sql <- SqlRender::render(sql,
                              cdm_database_schema = cdmDatabaseSchema,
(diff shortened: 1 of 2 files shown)


R/CohortConstruction.R
```diff
@@ -627,4 +629,25 @@ saveAndDropTempInclusionStatsTables <- function(connection,
...
+}
+
+
+.warnMismatchSqlInclusionStats <- function(sql) {
+  if (any(stringr::str_detect(string = sql, pattern = "_inclusion_result"), 
+          stringr::str_detect(string = sql, pattern = "_inclusion_stats"), 
+          stringr::str_detect(string = sql, pattern = "_summary_stats")
+          )
+  ) {
+    if (isFALSE(generateInclusionStats)) {
```
schuemie, line 641: Are you referring to a global variable (`generateInclusionStats`) here?!?

This should be an argument of the `.warnMismatchSqlInclusionStats` function

schuemie:
You did not run R check on this. I can tell, because you're failing the [automated R check on travis](https://travis-ci.com/github/OHDSI/CohortDiagnostics/builds/160902505). Please always run R check before making a pull request.

In this case the fail is because you're using stringr, but didn't declare that dependency in the DESCRIPTION file.

R/CohortConstruction.R
gowthamrao: Thank you !

gowthamrao: Thank you! 

```
.warnMismatchSqlInclusionStats: no visible binding for global variable
  â€˜generateInclusionStatsâ€™
launchCohortExplorer: no visible binding for global variable
  â€˜shinySettingsâ€™
Undefined global functions or variables:
  generateInclusionStats shinySettings
```

```
* checking dependencies in R code ... WARNING
'::' or ':::' import not declared from: â€˜stringrâ€™
```

Issue #32, opened 2020-04-07
Ability to pass FeatureExtraction::createCovariateSettings to getCohortCharacteristics


edward-burn:
At the moment, I believe getCohortCharacteristics uses createDefaultCovariateSettings() from FeatureExtraction as the default for covariateSettings for getCohortCharacteristics(). It would be great if it were also possible to have the ability to use FeatureExtraction::createCovariateSettings() here, so as to include features not included in the default set.

gowthamrao:
@schuemie  seems like a good enhancement to have. I have noticed that the default cohort characterization specification is not sufficient to feel confident in the diagnostics of a cohort.

e.g. if we are building the phenotype of 'joint swelling' - i would be interested in having detailed information in table 1 of concepts that may cause joint swelling e.g. Rheumatoid arthritis, trauma, knee infection, etc. Only Rheumatoid arthritis would be in default table 1.

schuemie:
So I think there are two issues:
1. The need to use different covariateSettings when building the cohort characterization
2. The need to show different concepts in the default 'table 1'.

For 2, a solution is to simple replace the [Table1Specs](https://github.com/OHDSI/CohortDiagnostics/blob/master/inst/shiny/DiagnosticsExplorer/Table1Specs.csv). The format of these specifications is taken from FeatureExtraction. It could use some documentation. Any volunteers?

Adding covariateSettings as an argument to [runCohortDiagnostics](https://github.com/OHDSI/CohortDiagnostics/blob/master/R/RunDiagnostics.R#L58) (defaulting to the default covariate settings) sounds fine to me. @gowthamrao : you want to take that on?

gowthamrao:
> For 2, a solution is to simple replace the Table1Specs. The format of these specifications is taken from FeatureExtraction. It could use some documentation. Any volunteers?

Happy to. I will create an issue in FeatureExtraction

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/commit/155b5bbaac3c90f00a52ae4f0924f877bf3af92f

PR #46, opened 2020-04-19
Adds ability to pass covariateSettings object to modify default table 1 specs

gowthamrao:
Addresses enhancement request here https://github.com/OHDSI/CohortDiagnostics/issues/32

### DIFF
--- a/R/RunDiagnostics.R
+++ b/R/RunDiagnostics.R
@@ -49,10 +49,9 @@
 #' @param runIncidenceRate            Generate and export the cohort incidence  rates?
 #' @param runCohortOverlap            Generate and export the cohort overlap?
 #' @param runCohortCharacterization   Generate and export the cohort characterization?
-#' @param covariateSettings           (OPTIONAL) Either an object of type \code{covariateSettings} as created using one of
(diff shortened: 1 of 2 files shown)


R/RunDiagnostics.R
```diff
@@ -49,6 +49,10 @@
 #' @param runIncidenceRate            Generate and export the cohort incidence  rates?
 #' @param runCohortOverlap            Generate and export the cohort overlap?
 #' @param runCohortCharacterization   Generate and export the cohort characterization?
+#' @param covariateSettings           (OPTIONAL) Either an object of type \code{covariateSettings} as created using one of
```
schuemie, line 52: This argument is not optional, it must be provided. Luckily, if the user doesn't specify it , a default value is used. (The same is true for other arguments like `runCohortOverlap`.)

Please don't document default values of arguments. These will be automatically shown in the help file, where people can see it defaults to `FeatureExtraction::createDefaultCovariateSettings()`.

gowthamrao: > Please don't document default values of arguments. These will be automatically shown in the help file, where people can see it defaults to FeatureExtraction::createDefaultCovariateSettings().

Thank you!

Issue #48, opened 2020-04-21
Drop ff in favor of Andromeda


schuemie:
Should be done in a separate 'andromeda' branch. Likely only requires changes in [this file](https://github.com/OHDSI/CohortDiagnostics/blob/master/R/CohortCharacterizationDiagnostics.R#L41). 

See [this post](https://forums.ohdsi.org/t/replacing-ff-with-rsqlite/9359/32).

schuemie:
@gowthamrao : did you test the Andromeda version of CohortDiagnostics with the Andromeda versions of DatabaseConnector and FeatureExtraction?

PR #49, opened 2020-04-22
Dropped ff in favor of Andromeda

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/issues/48

This PR has not been tested with R-check


DESCRIPTION
```diff
@@ -24,7 +24,9 @@ Imports:
 	FeatureExtraction,
 	ff,
```
schuemie, line 25: Would you do the honor of removing this dependency ;-)

gowthamrao: Done https://github.com/OHDSI/CohortDiagnostics/pull/49/commits/568851d9a9ac11fd11befbe65097f4200b875af9

Issue #52, opened 2020-04-23
Deduplicate cohorts in Shiny app


schuemie:
When CohortDiagnostics is run across both Windows and non-Windows machines, the cohort.csv files are different because one uses `\n` and the other uses `\r\n` as end-of-line. This causes duplication of the cohorts in the Shiny app.

One quick fix is to switch [this line](https://github.com/OHDSI/CohortDiagnostics/blob/master/inst/shiny/DiagnosticsExplorer/global.R#L63) with the next.

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/blob/e1cf412b3a9d56e069cbe2924250f1b8e2311573/inst/shiny/DiagnosticsExplorer/global.R#L63

PR #64, opened 2020-05-21
Deduplicate cohorts in Shiny app

gowthamrao:
Fixes https://github.com/OHDSI/CohortDiagnostics/issues/52


schuemie:
You really don't need dplyr here. Just move the unique command down two lines. 

The aim should always be to minimize the number of dependencies, but especially for Shiny apps since the dependencies are not automatically installed by the Shiny Server. Your action here would probably require everyone to send an e-mail to their server admin to install the additional dependency!

Issue #68, opened 2020-06-10
revise characterization features and time windows


pbr6cornell:
Characterization should use:

condition (not condition group)
drug exposure (not drug group)
procedure
measurement

time windows to compute should be:
-365 -> -31
-30 -> -1
0 -> 0
1 -> 30
31 -> 365

(this will allow characterization to be used explicitly to determine if the index date seems appropriate for a given cohort, or whether observations occuring before or after seem out-of-sequence with expectations.   verbatim concepts more useful than group'ed concepts for detecting anamolies)

gowthamrao:
Same as?

https://github.com/OHDSI/CohortDiagnostics/issues/61

pbr6cornell:
related yes.  but difference is now i'm assigning it to you to get it done :)

schuemie:
And I still oppose this functionality, for these reasons:

- Complexity. Yes, Patrick might now how to interpret this, but no one else will. Right now the characterization is consistent with the default used throughout HADES, so people might have a chance to get used to our crazy way of characterizing.

- Looking forward from the index date is always dangerous. Many bad causal inferences will be made.

schuemie:
Just to be clear: the question of whether the index date is correct is an important one. I just don't think the characterization diagnostic should be warped to answer that question. Another diagnostic should be developed, although I don't yet have a good idea what that would look like.

schuemie:
@pbr6cornell : would you mind providing some concrete examples of cohort issues that you would like CD to help identifying? I find it hard to think about this in the abstract.

gowthamrao:
@schuemie to understand if rate of certain feature (or feature cohort) go up or down in relation to a target cohort_ANY_date, I think OHDSI/Kala is the way to go. 

There is no need to break into arbitrary discrete periods

> time windows to compute should be:
-365 -> -31
-30 -> -1
0 -> 0
1 -> 30
31 -> 365

I can imagine x axis being time/date, and y axis being a measure of feature occurrence. And then we can empirically compare patterns of features in relation to cohort. We could even test it, like negative and positive control - and pick test that pattern.

gowthamrao:
@schuemie  @pbr6cornell  

Why cant this be solved with covariateSettings object of FeatureExtraction? Cohort Diagnostics does support passing covariateSettings in the call.

[...]

PR #93, opened 2020-07-13
Adds ability to generate and review temporal covariates

gowthamrao:


gowthamrao:
Resolves https://github.com/OHDSI/CohortDiagnostics/issues/68

R/RunDiagnostics.R
```diff
@@ -83,6 +99,8 @@ runCohortDiagnostics <- function(packageName = NULL,
                                  runCohortOverlap = TRUE,
                                  runCohortCharacterization = TRUE,
                                  covariateSettings = FeatureExtraction::createDefaultCovariateSettings(),
+                                 runTemporalCohortCharacterization = FALSE,
+                                 temporalCovariateSettings = NULL,
```
schuemie, line 103: Why not simply assign the temporal covariate settings you recommend? That way, you don't have to have to elaborate explanation above, because people could see what the default is. So:

```r
 runTemporalCohortCharacterization = FALSE,
 temporalCovariateSettings = FeatureExtraction::createTemporalCovariateSettings(useConditionOccurrence = TRUE,
                                                                                useConditionEraStart = TRUE,
etc.

File: inst/shiny/DiagnosticsExplorer/server.R [hunk 2]
```diff
@@ -441,7 +447,9 @@ shiny::shinyServer(function(input, output, session) {
     databaseIds <- unique(data$databaseId)
     
     if (input$charType == "Pretty") {
-      data <- merge(data, covariate)
+      data <- data %>% 
+        dplyr::left_join(y = covariate, by = c('covariateId')) %>%
```
schuemie, line 451: Just so you know, you don't need a vector here. This works too:
```r
by = "covariateId"
```

inst/shiny/DiagnosticsExplorer/server.R
```diff
@@ -542,6 +550,118 @@ shiny::shinyServer(function(input, output, session) {
...
   
+  output$temporalCharacterizationTable <- DT::renderDataTable({
+    data <- temporalCovariateValue %>% 
+            dplyr::filter(.data$cohortId == cohortId(),
+                          .data$databaseId == input$databases,
+                          .data$timeId == timeId())
+    data$cohortId <- NULL
+    databaseIds <- unique(data$databaseId)
+    
+    if (input$charTypeTemporal == "Pretty") {
```
schuemie, line 561: You should ask Patrick what he thinks, but I don't think a 'Pretty' view makes sense for temporal covariates, since we're not trying to get an overall picture of the cohort, but are looking for specific codes that are behaving differently than expected. Just a raw view, sorted by prevalence, would be most helpful.

schuemie:
Looks ok, but I would remove the 'Pretty' view as mentioned, and currently the raw view doesn't seem to work (in the link you sent). 

I would also still like to see the time-period-per-column view.

pbr6cornell:
I agree with @schuemie that 'pretty view' is not needed in this context.  For diagnostics, we want to allow for exploration of all concepts within defined time periods for specific purposes purposes, and dont want to confuse this with the goal of a proper characterization to report out as a 'pretty' report.

Issue #101, opened 2020-07-20
Error in parsing sql for concept set ids 


gowthamrao:
There might be an error (bug) here

https://github.com/OHDSI/CohortDiagnostics/blob/ddc3f23472b53c2baec7c4469bc20ea3dcc24f6c/R/RunDiagnostics.R#L553

This incorrectly returns TRUE
```
    if (any(!(conceptSetIds %in% conceptSets$conceptSetId)) ||
        any(!(conceptSets$conceptSetId %in% conceptSetIds))) {
      stop("Mismatch in concept set IDs between SQL and JSON for cohort ", cohort$cohortFullName)
    }
```

In this scenario

```
sqlParts <- "INSERT INTO #Codesets (codeset_id, concept_id)\r\nSELECT 7 as codeset_id, c.concept_id FROM (select distinct I.concept_id FROM\r\n( \r\n  select concept_id from @vocabulary_database_schema.CONCEPT where concept_id in (2006301,2785792,2785999,2785992,2786008)\r\n\r\n) I\r\n) C UNION ALL \r\nSELECT 8 as codeset_id, c.concept_id FROM (select distinct I.concept_id FROM\r\n( \r\n  select concept_id from @vocabulary_database_schema.CONCEPT where concept_id in (4208113)\r\nUNION  select c.concept_id\r\n  from @vocabulary_database_schema.CONCEPT c\r\n  join @vocabulary_database_schema.CONCEPT_ANCESTOR ca on c.concept_id = ca.descendant_concept_id\r\n  and ca.ancestor_concept_id in (4208113)\r\n  and c.invalid_reason is null\r\n\r\n) I\r\n) C\r\n"

conceptSetIds <- as.numeric(gsub(pattern = "^.*SELECT ([0-9]+) as codeset_id.*$", replacement = "\\1", x = sqlParts, ignore.case = TRUE))

```

conceptSetIds only has 8, no 7

gowthamrao:
Replace the original regular expression with

```
stringr::str_extract_all(string = sqlParts, 
                                              pattern = "SELECT ([0-9]+) as ", simplify = TRUE) %>% 
                     stringr::str_replace_all(string = ., pattern = "SELECT ", replacement = "") %>% 
                     stringr::str_replace_all(string = ., pattern = " as ", replacement = "") %>% 
                     stringr::str_squish(string = .) %>% 
                     stringr::str_trim(string = .) %>% 
                     utils::type.convert()
```

[...]

PR #102, opened 2020-07-21
Changed the way regular expressions parses SQL

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/issues/101

### DIFF
--- a/R/RunDiagnostics.R
+++ b/R/RunDiagnostics.R
@@ -566,7 +566,13 @@ getConceptSets <- function(cohorts) {
     
     sqlParts <- SqlRender::splitSql( gsub("with primary_events.*", "", cohort$sql))
     sqlParts <- sqlParts[-1]
-    conceptSetIds <- as.numeric(gsub("^.*SELECT ([0-9]+) as codeset_id.*$", "\\1", sqlParts, ignore.case = TRUE))
+    conceptSetIds <- stringr::str_extract_all(string = sqlParts, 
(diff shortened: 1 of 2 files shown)


R/RunDiagnostics.R
```diff
@@ -566,7 +566,13 @@ getConceptSets <- function(cohorts) {
     
     sqlParts <- SqlRender::splitSql( gsub("with primary_events.*", "", cohort$sql))
     sqlParts <- sqlParts[-1]
-    conceptSetIds <- as.numeric(gsub("^.*SELECT ([0-9]+) as codeset_id.*$", "\\1", sqlParts, ignore.case = TRUE))
+    conceptSetIds <- stringr::str_extract_all(string = sqlParts,
```
schuemie, line 569: I don't understand how this is an improvement. I find the new code much harder to read, it is longer, and
- Does not require 'codeset_id' to be present, so the extracted ID might be something else?
- Is not case-sensitive, so could more easily break down. 

Please do not simply convert any syntax you are unfamiliar with with tidyVerse. I recommend reading [a primer on RegEx](http://marvin.cs.uidaho.edu/Handouts/regex.html) to learn more about this powerful tool.

schuemie:
I guess there was a specific case that the RegEx didn't cover? What was it?

I recommend simply modifying the RegEx to cover the new case, rather than trying to avoid RegEx altogether.

Issue #106, opened 2020-07-22
Retain concept id in covariate file


gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/blob/ddc3f23472b53c2baec7c4469bc20ea3dcc24f6c/R/RunDiagnostics.R#L447

Issue #87, opened 2020-07-02
Make the name of the cohort prominent on top every page


gowthamrao:
The name of the cohort being reviewed in not prominently displayed. For every page in submenu of cohort diagnostics, we need to be explicit on the cohort. One idea is to put the cohort name - clearly on top (see yellow highlighted area)

![image](https://user-images.githubusercontent.com/13936600/86343868-81769480-bc27-11ea-984c-3fa1687ec16a.png)

Currently, the only way to know which cohort is being reviewed is to select the drop down (which is often truncated in long names)

![image](https://user-images.githubusercontent.com/13936600/86343963-a7039e00-bc27-11ea-8dac-6c6793ba07c0.png)

and when i select the drop down to reveal the full cohort name - the experience is 'jumpy' (especially when we have a large number of cohorts to review).

![image](https://user-images.githubusercontent.com/13936600/86344264-1083ac80-bc28-11ea-87e2-c8d7d9726437.png)

Issue #61, opened 2020-05-21
Change default lookback periods for cohort characteristics diagnostics


gowthamrao:
Purpose of `cohort characteristics diagnostics` is to get an empirical measure of certain selected features in relation to cohort_start_date. The `diagnosis` here is the empirical affirmation of investigators confidence that a certain selected feature that is apriori expected to co-occur/associated with cohort - is observed in sufficient numbers. This metric may be used to iteratively improve a phenotype algorithm - where algorithms maybe iterated till the investigator is satisfied with the observed numbers of the apriori selected feature. 

In current implementation of the characteristics diagnostics  
- the features are defaults from featureExtraction
- the time horizons are the default time horizons from feature extraction.

More details here  https://github.com/OHDSI/CohortDiagnostics/issues/32

Having look-back of day -365 to day 0, is generally not informative for 'diagnostics; This default is useful for cohort characterization - where we are trying to infer new characteristics about a cohort - understanding the characteristics of a cohort is different from `diagnostics` where we trying to infer the performance of a phenotype algorithm.

To understand the performance of a phenotype algorithm it is more informative to understand the features of a cohort that is closer to cohort_start_date i.e. events that happened immediately before, on and immediately after cohort entry event - are likely to help us `diagnose` the performance of an algorithm. Thus we would like the default look back periods to be -30 to -1/ 0 days /+1 to +30 days.

@pbr6cornell have we captured our insight during [Scylla](https://www.ohdsi.org/wp-content/uploads/2020/05/OHDSI-SCYLLA-PLE-COVID-effectiveness-protocol-v1.0-final-with-appendix.pdf) project accurately?

Also related to https://github.com/OHDSI/CohortDiagnostics/issues/32
https://github.com/OHDSI/FeatureExtraction/issues/92

Tagging @schuemie

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/commit/155b5bbaac3c90f00a52ae4f0924f877bf3af92f

Issue #26, opened 2020-03-31
add Concept ID column to raw view of cohort characterization tables


jamieweaver:
Adding Concept ID column to raw view of cohort characterization tables will facilitate using concepts in development activities supported by CohortDiagnostics results

PR #107, opened 2020-07-22
Temporal characterization and bug fixes

gowthamrao:
Resolves https://github.com/OHDSI/CohortDiagnostics/issues/106
https://github.com/OHDSI/CohortDiagnostics/issues/87
https://github.com/OHDSI/CohortDiagnostics/issues/68
https://github.com/OHDSI/CohortDiagnostics/issues/61
https://github.com/OHDSI/CohortDiagnostics/issues/52
https://github.com/OHDSI/CohortDiagnostics/issues/26

### DIFF
--- a/R/CohortCharacterizationDiagnostics.R
+++ b/R/CohortCharacterizationDiagnostics.R
@@ -46,14 +46,14 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
                                      cohortTable = "cohort",
                                      cohortId,
                                      covariateSettings) {
-
+  
   start <- Sys.time()
-
+  
   if (is.null(connection)) {
     connection <- DatabaseConnector::connect(connectionDetails)
     on.exit(DatabaseConnector::disconnect(connection))
   }
-
+  
   if (!checkIfCohortInstantiated(connection = connection,
                                  cohortDatabaseSchema = cohortDatabaseSchema,
                                  cohortTable = cohortTable,
@@ -65,7 +65,7 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
                                   attr(delta, "units")))
     return(data.frame())
   }
-
+  
   data <- FeatureExtraction::getDbCovariateData(connection = connection,
                                                 oracleTempSchema = oracleTempSchema,
                                                 cdmDatabaseSchema = cdmDatabaseSchema,
@@ -74,7 +74,7 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
                                                 cohortId = cohortId,
                                                 covariateSettings = covariateSettings,
                                                 aggregated = TRUE)
-
+  
   result <- tidyr::tibble()
   if (!is.null(data$covariates)) {
     n <- attr(x = data, which = "metaData")$populationSize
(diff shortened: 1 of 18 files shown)


R/CohortCharacterizationDiagnostics.R
```diff
@@ -45,7 +45,7 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
                                      cohortDatabaseSchema = cdmDatabaseSchema,
                                      cohortTable = "cohort",
                                      cohortId,
-                                     covariateSettings = FeatureExtraction::createDefaultCovariateSettings()) {
+                                     covariateSettings) {
```
schuemie, line 48: Just curious: why no longer provide a default?

R/RunDiagnostics.R
```diff
@@ -83,6 +99,8 @@ runCohortDiagnostics <- function(packageName = NULL,
                                  runCohortOverlap = TRUE,
                                  runCohortCharacterization = TRUE,
                                  covariateSettings = FeatureExtraction::createDefaultCovariateSettings(),
+                                 runTemporalCohortCharacterization = FALSE,
+                                 temporalCovariateSettings = NULL,
```
schuemie, line 103: As I mentioned before, I would here just assign the covariate settings you describe in the documentation. That makes it more transparent what happens.

R/RunDiagnostics.R
```diff
@@ -461,34 +479,45 @@ runCohortDiagnostics <- function(packageName = NULL,
                                          cohortTable = cohortTable,
```
schuemie, line 479: I don't know if it something weird with GitHub, but it seems these 3 lines should be deleted?

schuemie:
Looks good as far as I can tell. I had some minor comments.

R/CohortCharacterizationDiagnostics.R
gowthamrao: When i added temporalCharacterization, i made this function `getCohortCharacteristics` support both temporal and non temporal characterization. 

I might be wrong here, but i think, i did not want to default it to non temporal covariate settings.

thinking back - i could have left it as is

R/RunDiagnostics.R
gowthamrao: Sorry. i think i am not following, and I know you are right..  
To make it transparent, i had added this to the documentation
https://github.com/OHDSI/CohortDiagnostics/blob/ec4cf215144efe41a3de52bac7d3ae2e5228570d/R/RunDiagnostics.R#L26

but i think you are saying, there is a better way

R/RunDiagnostics.R
gowthamrao: i cant see it on local copy....but i will make some style changes to removes some empty lines

R/RunDiagnostics.R
schuemie: Yes, why not assign the default value, instead of making it so a NULL value mysteriously leads to the default value? So change line 103 to:
```r
temporalCovariateSettings = FeatureExtraction::createTemporalCovariateSettings(useConditionOccurrence = TRUE,
                                                                                 useConditionEraStart = TRUE,
                                                                                 useDrugEraStart = TRUE,
                                                                                 useProcedureOccurrence = TRUE,
                                                                                 useMeasurement = TRUE,
                                                                                 useObservation = TRUE,
                                                                                 temporalStartDays = c(-365,-30,0,1,31),
                                                                                 temporalEndDays = c(-31,-1,0,30,365)),
```

schuemie:
Ok

R/RunDiagnostics.R
gowthamrao: Got it! thanks for your guidance!

PR #141, opened 2020-07-30
Release version 1.1

gowthamrao:


schuemie:
As I mentioned, I was still performing some tests I'd like to see completed before releasing version 1.1.0. I found the following issues:

1. The new temporal characterization requires the new version of FeatureExtraction I just released (v3.0.1). This should be recorded in the DESCRIPTION, or else most users will experience weird error messages: https://github.com/OHDSI/CohortDiagnostics/issues/142

2. The Shiny app now requires shinyWidgets to be installed, but this is not automatically installed (unlike other Shiny app dependencies): https://github.com/OHDSI/CohortDiagnostics/issues/143

I recommend rolling back the 1.1.0 release, make these fixes, and release again.

schuemie:
Also requires version of SqlRender that hasn't been released yet. Will release shortly.

PR #168, opened 2020-08-12
Cohort filtering - fixes #167

### LINKED ISSUE #167: Cohort construction fails when cohortIds specified for subset
The logic [in the code here](https://github.com/OHDSI/CohortDiagnostics/blob/master/R/CohortConstruction.R#L33-L34) appears to be the opposite of the desired intent which is to filter the `cohorts` based on the `cohortIds` specified by the function parameter. Additionally, the [cohortId function parameter](https://github.com/OHDSI/CohortDiagnostics/blob/master/R/CohortConstruction.R#L19) seems to use a global variable instead of assigning a default that is either hard-coded or based on another parameter. 

I'll also check the other methods in the CohortConstruction.R file as they appear to have the same behavior.

anthonysena:
Aims to fix #167


R/CohortConstruction.R
```diff
@@ -30,8 +30,10 @@ getCohortsJsonAndSqlFromPackage <- function(packageName = packageName,
                               extension = "csv", 
                               add = errorMessage)
   
-  cohorts <- readr::read_csv(pathToCsv, col_types = readr::cols()) %>%
-    dplyr::filter(!.data$cohortId %in% cohortIds)
+  cohorts <- readr::read_csv(pathToCsv, col_types = readr::cols())
+  if (!is.null(cohortIds)) {
+    cohorts <- cohorts %>% dplyr::filter(.data$cohortId %in% cohortIds)
```
anthonysena, line 35: Note that the `filter` call is using a different logic:

Original: dplyr::filter(**!**.data$cohortId %in% cohortIds)
Revised: dplyr::filter(.data$cohortId %in% cohortIds)

gowthamrao: Thanks for the fix! @anthonysena

PR #175, opened 2020-08-13
Fixes #174

### LINKED ISSUE #174: Ensure concept sets across cohort definitions are unique
Facing an issue whereby a list of cohort definitions have the same concept set expression but different concept set IDs is causing the orphan concept checker in Cohort Diagnostics to fail. Reason being is that while the concept set expression is unique, the resulting SQL may differ since the conceptSetId is included as part of the SQL. Tracing through the code, this appears to be the offending code:

https://github.com/OHDSI/CohortDiagnostics/blob/master/R/ConceptSets.R#L239-L241 

@gowthamrao per our discussion, this will require some revision.

anthonysena:
Fixes #174

### DIFF
--- a/R/ConceptSets.R
+++ b/R/ConceptSets.R
@@ -156,7 +156,7 @@ compareAndCombineConceptSetsDefinitionForCohorts <-
     }
     conceptSets <- dplyr::bind_rows(conceptSets) %>%
       dplyr::arrange("cohortId", "conceptSetId")
-    
+
     uniqueConceptSets <- conceptSets %>%
       dplyr::select(.data$conceptSetExpression) %>%
       dplyr::distinct() %>%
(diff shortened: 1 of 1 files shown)


R/ConceptSets.R
```diff
@@ -237,10 +237,10 @@ runConceptSetDiagnostics <- function(connection,
   
   conceptSets <- compareAndCombineConceptSetsDefinitionForCohorts(subset)
   uniqueConceptSets <- conceptSets %>% 
-                        dplyr::select(.data$uniqueConceptSetId, .data$conceptSetName, .data$conceptSetSql) %>% 
-                        dplyr::distinct()
-  instantiateUniqueConceptSets(cohorts = subset,
```
anthonysena, line 242: I also removed the `cohorts = subset` parameter here since this appears to have been removed and was causing an error.

PR #196, opened 2020-08-28
Added a deployment guide for using renvs to vignette

azimov:
People have had trouble deploying the application to shared shiny servers so I wrote this guide as a starting point to help people understand what to do if they're struggling with system packages.


gowthamrao:
Reopening and targeting version 2

PR #253, opened 2020-09-21
flexible schema binding variable for DDL generation; script alignment; backup database script

jdide:


R/ResultsDataModel.R
```diff
@@ -21,12 +21,14 @@
...
 #' @param packageName    The name of the R package whose output model we are documenting.
+#' @param schemaBindingVariable     The name of the schema binding parameter
 #' 
 #' @export
 createDdl <- function(packageName,
                       packageVersion,
                       modelVersion,
-                      specification){
+                      specification,
+                      schemaBindingParam = "databaseSchema"){
```
gowthamrao, line 31: should be `schemaBindingVariable = "databaseSchema"` based on documenation?

gowthamrao: We could call it is as 'databaseSchema' to represent a generic database + schema combination

gowthamrao (changes requested):
Please see request for small changes

R/ResultsDataModel.R
jdide: > should be `schemaBindingVariable = "databaseSchema"` based on documenation?

good catch @gowthamrao, can you please review again?

PR #279, opened 2020-10-01
Connection pooling

jdide:


inst/shiny/DiagnosticsExplorer/global.R
```diff
@@ -1,59 +1,92 @@
...
+    }
+  }
+})
+
+
+if (is.null(shinySettings$connectionDetails)) {
+  warning("No database connection details. Looking for local data.")
+  if (is.null(shinySettings$dataFolder)) {
+    stop("No mechanism to load data.")
+  } else {
```
azimov, line 57: Else logic after stop is redundant so this could be tidied up

azimov (approved):

inst/shiny/DiagnosticsExplorer/global.R
jdide: thanks @azimov code cleaned up. can you please review again?

PR #319, opened 2021-02-19
Add github actions continuous integration

azimov:
Updated to use github actions and removed old travis configuration.


schuemie (approved):

.github/workflows/R_CMD_check_Hades.yml
```diff
@@ -0,0 +1,193 @@
...
+
+      - name: Install libssh
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get install libssh-dev
+      - name: Install dependencies
+        run: |
+          remotes::install_deps(dependencies = TRUE, INSTALL_opts=c("--no-multiarch"))
+          remotes::install_cran("rcmdcheck")
+          remotes::install_github(c("ohdsi/Eunomia", "ohdsi/FeatureExtraction", "ohdsi/ROhdsiWebApi", "ohdsi/CirceR"), dependencies = TRUE)  # Note that this is not a long term solution but a work around to see if the rest of the build works
```
azimov, line 89: @schuemie this is my attempt to fix an error caused by use of "Remotes:" in the DESCRIPTION file. However, if the package still requires these github versions maybe its best to remove `--as-cran` from `rcmdcheck` args?

schuemie: Yes, please remove --as-cran. We like the Remotes section in our DESCRIPTION files (it means install_github() automatically installs these dependencies if needed).

.github/workflows/R_CMD_check_Hades.yml
```diff
@@ -0,0 +1,192 @@
...
+
+      - name: Remove check folder if exists
+        if: runner.os == 'macOS'
+        run: unlink("check", recursive = TRUE)
+        shell: Rscript {0}
+
+      - name: Check
+        env:
+          _R_CHECK_CRAN_INCOMING_REMOTE_: false
+        run: rcmdcheck::rcmdcheck(args = c("--no-manual", "--no-multiarch"), error_on = "error", check_dir = "check")
```
azimov, line 105: I changed `error_on` param from warning to error to confirm the checks work. This should probably be changed back before a release.

PR #328, opened 2021-03-24
use DatabaseConnector from CRAN

### LINKED ISSUE #327: ohdsi/DatabaseConnector should be loaded from CRAN instead
If you install HADES and already have DatabaseConnector installed, then this package will cause your system to download the package from GitHub even if it is already installed. Please remove the dependency and use the one from CRAN instead.

sdebruyn:
closes #327


sdebruyn:
@gowthamrao done

gowthamrao (approved):

Issue #420, opened 2021-05-27
Optimize cohort overlap so that we may do all by all in one query


gowthamrao:
Right now - cohort overlap performs computation on any two cohort combination at a time.

gowthamrao:
@azimov  here is a real comparison

older method - comparing n = 200 cohorts

> Running Cohort Overlap took 6.69 hours

New optimized method
> Running Cohort Overlap took 6.1 mins

Thats a pretty good improvement dont you think :)

PR #421, opened 2021-05-27
Cohort overlap optimization

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/issues/420

Hey @azimov cohort overlap compares every cohort combination. Current implementation runs one query for every combination of n*(n-1) of cohorts. This has become bottle neck - as most of time is just I/O (99%).

Maybe this will help speed up?

### DIFF
--- a/R/CohortComparisonDiagnostics.R
+++ b/R/CohortComparisonDiagnostics.R
@@ -21,56 +21,14 @@ computeCohortOverlap <- function(connectionDetails = NULL,
                                  targetCohortIds,
                                  comparatorCohortIds,
                                  batchSize = 200) {
-  start <- Sys.time()
+  startTime <- Sys.time()
   
   if (is.null(connection)) {
     connection <- DatabaseConnector::connect(connectionDetails)
     on.exit(DatabaseConnector::disconnect(connection))
   }
   
-  ## commenting out checks if cohort is instantiated, as this is not essential.
-  ## if a cohort has 0 records, the overlap query will return NULL
-  # if (!checkIfCohortInstantiated(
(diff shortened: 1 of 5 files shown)


R/CohortComparisonDiagnostics.R
```diff
@@ -18,75 +18,97 @@ computeCohortOverlap <- function(connectionDetails = NULL,
...
-    )
-    delta <- Sys.time() - start
-    ParallelLogger::logInfo(paste(
-      "Computing overlap took",
-      signif(delta, 3),
-      attr(delta, "units")
-    ))
-    return(tidyr::tibble())
-  }
+  ## commenting out checks if cohort is instantiated, as this is not essential.
```
azimov, line 31: Code should either be deleted or kept, we can let git manage the history :)

azimov (commented):
This may be better but its hard to tell, can you provide a benchmark time of this vs the time taken doing the overlap in pairs? The code of only using a single SQL statement is more elegant. 

Also, it looks likes the tests haven't ran in codecov for some reason so we need to verify the new code produces the same outputs as before.

R/CohortComparisonDiagnostics.R
gowthamrao: good point!

gowthamrao:
@azimov 

On eunomia dataset

Running Cohort Overlap took 6.21 secs without optimization
Running Cohort Overlap took 5.12 secs

we have to verify if the sql is accurate

gowthamrao:
I will rewrite the SQL, I don't think it's correct

Issue #434, opened 2021-06-04
Expose cohort diagnostics data model via R functions


gowthamrao:
Add functions to package to expose results from data model.
Support both db and premerged file
Write script in package maintenance that copies the functions into shiny

gowthamrao:
Implemented for next version - in develop branch

PR #440, opened 2021-06-07
Creating a pattern to make the data in cohort diagnostics results data model easily callable

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/issues/434

The same script is intended to be copied from the base package into DiagnosticsExplorer during package maintenance


gowthamrao:
@azimov  what do you think of this pattern to expose results

azimov:
This is a step in the right direction, I like that the model code is part of the package but I think this will have some limitations on data.ohdsi.org as the CohortDiagnostics package will have to be installed. Allowing this would be my preference as we could easily make the code here in the package. @schuemie you had opinions that were agains

Because the functions here are in the package I would also like to see unit tests for them that ensure that they work in a universal manner.
This pattern is much more testable than the code being purely inside the shiny app so thats a good thing.

My longer term preference would be a pattern that uses model classes (either S4, R6 or reference classes). This way we could define a set of required methods (e.g. an interface) that the shiny model would always expect. I implemented something along these lines in a recent package, CEMConnector, [here](https://github.com/OHDSI/CEMConnector/blob/main/R/DatabaseConnections.R) and [here](https://github.com/OHDSI/CEMConnector/blob/main/R/Backend.R) with R6 classes.

R/ResultsDataModelQueries.R
```diff
@@ -0,0 +1,414 @@
...
+  for (letter in letters) {
+    string <-
+      gsub(paste("_", letter, sep = ""), toupper(letter), string)
+  }
+  string <- gsub("_([0-9])", "\\1", string)
+  return(string)
+}
+
+# private function - not exported
+camelCaseToSnakeCase <- function(string) {
```
azimov, line 41: Can these methods not be used directly from SqlRender? This is a dependency of the package.

gowthamrao: Yup - it can be - but note prior comment to explain why.

gowthamrao:
>  I think this will have some limitations on data.ohdsi.org as the CohortDiagnostics package will have to be installed.

This design does **not** introduce dependency for cohort diagnostics in shiny. The entire 'ResultsDataModelQueries.R' script is copied over to shiny during package maintenance. 

https://github.com/OHDSI/CohortDiagnostics/blob/f04bf2472494a3a3c394a21402b8bcb247b614b8/extras/PackageMaintenance.R#L77

That way - we dont have a dependency. Yes, there is code duplication - but the trade off is removing the dependency. This is @schuemie  trick :)

gowthamrao:
> My longer term preference would be a pattern 

Yup - you will be writing that :)

gowthamrao:
Based on feedback from @schuemie  i am going to redo this. We will use the dataSource approach  and one function for each type of diagnostics being exposed.

that way - if a user writes a script for one data source e.g. db, the same script may be used to create a report using premerged file - with only change being the 'data source pointer'

PR #441, opened 2021-06-08
Export functions2

gowthamrao:
@azimov  @schuemie  here is the new pattern


gowthamrao:
replaced https://github.com/OHDSI/CohortDiagnostics/pull/440

inst/shiny/DiagnosticsExplorer/R/Results.R
```diff
@@ -1,88 +1,57 @@
...
-              resultsDatabaseSchema = resultsDatabaseSchema,
-              vocabularyDatabaseSchema = vocabularyDatabaseSchema))
-}
-
-
-createFileDataSource <- function(premergedDataFile, envir = new.env()) {
-  load(premergedDataFile, envir = envir)
-  return(envir)
-}
+# createDatabaseDataSource <- function(connection,
```
azimov, line 1: its always best to remove code rather than commenting it. (makes the reviewers job easier ;))

R/ResultsDataModelQueries.R
```diff
@@ -0,0 +1,294 @@
...
+          sql = sql,
+          ...,
+          snakeCaseToCamelCase = snakeCaseToCamelCase
+        ) %>% dplyr::tibble()
+      )
+    }
+  }
+
+# private function - not exported
+getDataForDatabaseIdsCohortIds <- function(dataSource,
```
azimov, line 158: This function has a lot of flow control so really needs some careful unit tests.

R/ResultsDataModelQueries.R
```diff
@@ -0,0 +1,294 @@
...
+                                           dataTableName) {
+  if (any(is.null(databaseIds), length(databaseIds) == 0)) {
+    stop("Database Ids not correctly specified.")
+  }
+  
+  if (any(!exists('cohortIds'), length(cohortIds) == 0)) {
+    stop("Cohort Ids not correctly specified.")
+  }
+  
+  if (is(dataSource, "environment")) {
```
azimov, line 170: Instead of environment a better option may be to give the dataSource an S3 class. This would be quite flexible and could be used to improve performance in the future as there maybe sections of code that would return joins from the database, rather than pulling out the table and doing it in dplyr.

For example:

```
getDataForDatabaseIdsCohortIds <- function(dataSource,...) {
    UseMethod("getDatabaseIdsCohortIds", dataSource)
}

getDatabaseCohortIds.environmentDataModel <- function(...) {
   .... # Getting data from premerged file
}

getDatabaseIdsCohortIds.PooledConnection <- function(...) {
   .... # However databases should be queried if they're pooled connections
}

x <- environment()
readRDS("PreMergedFile.rds", envir = x) 
class(x) <- "environmentDataModel"


x <- DatabaseConnector::connection(...)
```

azimov:
Overall this is good but there are two things that are needed:
* Unit tests for the data model - if any new code is in the package folder the rule should be to have test coverage for it!
*  Fix to build, it looks like `createDatabaseDataSource` function is missing from the results

I also think this could really benefit from using S3 classes. It would be a relatively minor change now but could mean a huge amount of flexibility in the future.

R/ResultsDataModelQueries.R
gowthamrao: Interesting proposal - maybe next version? Can we make this an issue for future work?

schuemie: The main reason we went with an environment is because our Shiny apps loaded everything in the global environment, and you could just point to the global environment in these functions. Using S3 classes may be cleaner (but will require some rewrite of the Shiny app).

R/ResultsDataModelQueries.R
```diff
@@ -0,0 +1,977 @@
...
+#'
+#' @export
+getConceptDetails <- function(dataSource = .GlobalEnv,
+                              vocabularyDatabaseSchema = NULL,
+                              conceptIds) {
+  table <- 'concept'
+  if (!is.null(vocabularyDatabaseSchema) &&
+      is(dataSource, "environment")) {
+    warning(
+      'vocabularyDatabaseSchema provided for function getConceptDetails in non database mode. This will be ignored.'
```
schuemie, line 431: Would you mind consistently using double quotes in your R code?

man-roxygen/DataSource.R
```diff
@@ -0,0 +1 @@
+#' @param dataSource           A list object that is the output of \code{createDatabaseDataSource} function.
```
schuemie, line 1: Need to be a bit more verbose here. This would be the place to mention that the data extraction functions can work with a database of a set of files. It would also be good to mention the createFileDataSource function as an alternative to createDatabaseDataSource. 

The notion of dataSources will be tricky for many users, so you'll need to help them a little bit.

schuemie:
You seem to be missing some documentation, which is why the builds are failing. Other than that I think you're in good shape (see my minor comments).

R/ResultsDataModelQueries.R
gowthamrao: Ok then definitely future version

R/ResultsDataModelQueries.R
gowthamrao: absolutely

man-roxygen/DataSource.R
gowthamrao: done

azimov:
The code looks good now but I'm not sure why your test cases for the data model aren't showing up in the coverage reports.

schuemie:
Quick tip: If you for reason find that you need to do another commit and push within minutes of the previous one, you can cancel the previous GitHub Action to save time: https://github.com/OHDSI/CohortDiagnostics/actions (there was a long queue of runs. I killed all but the last)

PR #446, opened 2021-06-17
Fixed bugs related to build

gowthamrao:


inst/sql/sql_server/CohortOverlap.sql
```diff
@@ -147,138 +147,138 @@ GROUP BY all1.target_cohort_id,
...
 	value
 INTO #cohort_overlap_long
 FROM (
-	SELECT DISTINCT target_cohort_id cohort_id,
+	SELECT DISTINCT target_cohort_id AS cohort_id,
 		comparator_cohort_id,
-		'es' attribute,
-		'o' attribute_type, -- overlap
-		either_subjects value
+		'es'  AS attribute_name,
```
azimov, line 157: @gowthamrao  You were right before, i think we want to remove the as keyword as "attribute_name" should fix the error on postgres

gowthamrao: i think its ok to use for column alias, not for table alias
probably best practice to use it as column alias

Issue #481, opened 2021-07-15
Package tests are failing because SqlRender cannot locate package SQL files


keesvanbochove:
When testing the package using `devtools::test()`, I get the following error message:
```
Error: Cannot find 'CreateConceptIdTable.sql' in the 'sql/sql_server' folder of the 'CohortDiagnostics' package.
Backtrace:
 1. base::system.time(...) test-againstEunomia.R:49:2
 2. CohortDiagnostics::runCohortDiagnostics(...)
 3. SqlRender::loadRenderTranslateSql(...) /root/CohortDiagnostics/R/RunDiagnostics.R:390:2
```
This seems to be a problem where SqlRender cannot read the /sql folder from either the package /inst directory or directly from /sql in the package folder. So far, I've tested this in R 4.1 on Windows 64-bit and on Debian using a rocker Docker image, but I get this error in both cases. I also get it on both the master and testathon branches.

This problem then leads to several other checks failing, which makes sense since the cohorts are not created:
```
Error (test-againstCDM.R:124:3): Retrieve results from premerged file
Error (test-againstCDM.R:265:3): Create and upload results to results data model
Error (test-againstCDM.R:295:3): Retrieve results from remote database
Error (test-againstCDM.R:434:7): Data removal works
Error: Error executing SQL:
org.postgresql.util.PSQLException: ERROR: syntax error at or near "."
  Position: 22
```

Running `R CMD check` instead fails on this
```
* checking for file 'CohortDiagnostics/DESCRIPTION' ... ERROR
Required field missing or empty:
  'Author'
* DONE
```
but after correcting for that, I get the same errors as above (by the looks of it, it only outputs the text of the latest error to the console, and it reports the same end result `[ FAIL 6 | WARN 2 | SKIP 0 | PASS 29 ]` in current tip of `testathon` branch.

Any ideas on how to fix this? I've installed all required and recommended packages (R CMD check also checks this). Since I ran this into a vanilla R installation in Docker which I can easily recreate, I can reproduce the problem if needed.

keesvanbochove:
If it helps, here is `sessionInfo()` after running `devtools::test` (on Linux):
```
R version 4.1.0 (2021-05-18)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.2 LTS

Matrix products: default
BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.8.so

[...]

PR #482, opened 2021-07-16
Improve checks and usage of environment variables

keesvanbochove:
For a new user coming into testing of this package, it's tricky to get the environment variables right. This adds checks to see if the expected environmental variables exist, and also hard codes the cohort_diagnostics schema name as Martijn suggested in #481

### DIFF
--- a/extras/PackageMaintenance.R
+++ b/extras/PackageMaintenance.R
@@ -82,7 +82,7 @@ connectionDetails <- DatabaseConnector::createConnectionDetails(dbms = "postgres
                                                                 user = Sys.getenv("CDM5_POSTGRESQL_USER"),
                                                                 password = URLdecode(Sys.getenv("CDM5_POSTGRESQL_PASSWORD")),
                                                                 server = Sys.getenv("CDM5_POSTGRESQL_SERVER"))
-cohortDiagnosticsSchema <- "cohort_diagnostics"
+cohortDiagnosticsSchema <- Sys.getenv("CDM5_POSTGRESQL_COHORT_DIAGNOSTICS_SCHEMA")
 connection <- DatabaseConnector::connect(connectionDetails)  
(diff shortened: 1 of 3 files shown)


extras/PackageMaintenance.R
```diff
@@ -82,7 +82,7 @@ connectionDetails <- DatabaseConnector::createConnectionDetails(dbms = "postgres
                                                                 user = Sys.getenv("CDM5_POSTGRESQL_USER"),
                                                                 password = URLdecode(Sys.getenv("CDM5_POSTGRESQL_PASSWORD")),
                                                                 server = Sys.getenv("CDM5_POSTGRESQL_SERVER"))
-cohortDiagnosticsSchema <- Sys.getenv("CDM5_POSTGRESQL_COHORT_DIAGNOSTICS_SCHEMA")
+cohortDiagnosticsSchema <- "cohort_diagnostics"
```
gowthamrao, line 85: @keesvanbochove  why not check if the environment variable exists. If it returns NULL then pass the hardcoded value as a fall back default value?

tests/testthat/setup.R
```diff
@@ -26,9 +45,9 @@ if (runDatabaseTests) {
   
   cdmDatabaseSchema <- "eunomia"
   vocabularyDatabaseSchema <- "eunomia"
-  cohortDiagnosticsSchema <- Sys.getenv("CDM5_POSTGRESQL_COHORT_DIAGNOSTICS_SCHEMA")
+  cohortDiagnosticsSchema <- "cohort_diagnostics"
```
gowthamrao, line 48: see comment above - please change from hard coded

gowthamrao (changes requested):
@keesvanbochove  could you make these minor changes please

keesvanbochove:
OK, I see you made those too in the last hour, I've merged those in. Almost the same but I had forgotten to update PackageMaintenance.R.

PR #512, opened 2021-08-14
Version 2.1.1 hotfix 

gowthamrao:


gowthamrao:
@msuchard here is the 2.1.1 that should (hopefully) fix the VA issue. Github actions based tests are currently failing @azimov do you know if there has been a solution for this

PR #614, opened 2021-10-12
Upper limit on #calendar_years range for SQL compatibility

### LINKED ISSUE #613: Opaque error when running cohort diagnostitcs due to large observation_period_end_date in CDM
Hello, recently when running Pioneer, I ran into an opaque error, which propagated as a warning but appeared to be non-fatal bubbling up from cohort diagnostics. https://github.com/ohdsi-studies/PioneerWatchfulWaiting/issues/79

It looks like this stems from a few rows in our CDM which have an end_date of 9999 in our CDM, flowing over from some max date value that got populated somewhere in source data.

I know it's probably not recommended to have dates of 9999 as an end_date in the CDM, but it might be good to be defensive against this.

Can #calendar_years be updated to only contain the years relevant to the cohort? Or maybe this edge case can be dealt with in ComputeIncidenceRates? (the issue comes when we try to add 1 to the year 9999)

https://github.com/OHDSI/CohortDiagnostics/blob/2963a5db6b301ec8fd93fa3698cf1f1a5bfcf9ac/inst/sql/sql_server/GetCalendarYearRange.sql#L3-L5

https://github.com/OHDSI/CohortDiagnostics/blob/93e8915b504318300c68a565945dd17807a88567/R/IncidenceRates.R#L69-L80

https://github.com/OHDSI/CohortDiagnostics/blob/2963a5db6b301ec8fd93fa3698cf1f1a5bfcf9ac/inst/sql/sql_server/ComputeIncidenceRates.sql#L74

Jake-Gillberg:
Fixes #613

### DIFF
--- a/R/IncidenceRates.R
+++ b/R/IncidenceRates.R
@@ -67,7 +67,7 @@ getIncidenceRate <- function(connectionDetails = NULL,
     DatabaseConnector::querySql(connection, sql, snakeCaseToCamelCase = TRUE)
   
   calendarYears <-
-    dplyr::tibble(calendarYear = seq(yearRange$startYear, yearRange$endYear, by = 1))
+    dplyr::tibble(calendarYear = seq(yearRange$startYear, min(yearRange$endYear, 9998), by = 1))
   DatabaseConnector::insertTable(
     connection = connection,
     tableName = "#calendar_years",


azimov:
@gowthamrao this PR has been left open for a while, can we close it out or include it in 3.0.0 release?

Issue #677, opened 2021-11-12
Remove dependencies on WebAPI


azimov:
The use of webapi to insert cohort references is complicated and creates a huge number or different inputs to the package that are conflicting. We should remove ROhdsiWebApi as a dependency of this package - it may be the way that you generate a study package or get the cohort inputs, but this package should accept a common set of definitions and references.

Instead, I propose that this export should be a function of WebApi/ROhdsiWebApi -  e.g. `exportCohortDefinitionSet` which is just a name and set of cohort definitions ids. This can then be used by cohort generator and/or study packages. Any calls to create sql could be replaced with Circe.

gowthamrao:
Are we referring to https://github.com/OHDSI/CohortDiagnostics/blob/2963a5db6b301ec8fd93fa3698cf1f1a5bfcf9ac/R/CohortConstruction.R#L190

why is that an issue? btw - this is a functionality that we have had from day 1 - this is something that we dont like but kept it in for legacy reasons (i have always wanted to remove it)

we should be running cohort diagnostics in package mode - thats well developed. See this code
https://github.com/OHDSI/CohortDiagnostics/blob/master/extras/ExampleOfHowToCreateStudyPackageWithHydra.R

you can get running in full package mode in a few minutes - but we cant officially do this until Hydra develop is merged into master. ie Hydra has had Cohort Diagnostics skeleton in develop for months but has not been merged to master - so we cant use this (officially) yet.

gowthamrao:
Are we referring to https://github.com/OHDSI/CohortDiagnostics/blob/2963a5db6b301ec8fd93fa3698cf1f1a5bfcf9ac/R/CohortConstruction.R#L190

why is that an issue? If it's because of additional ways the package maybe called I agree that's bad.

 btw - this is a functionality that we have had for long time, this is something that we dont like but kept it in for legacy reasons (i have always advocated to remove it)

we should be running cohort diagnostics in package mode - thats well developed. See this code
https://github.com/OHDSI/CohortDiagnostics/blob/master/extras/ExampleOfHowToCreateStudyPackageWithHydra.R

you can get running in full package mode in a few minutes - but we cant officially do this until Hydra develop is merged into master. ie Hydra has had Cohort Diagnostics skeleton in develop.

Agree with Circe for SQL, but I think SQL should be created once during package creation and stored in package 

Yes, ideally cohort generator package should handle all that

gowthamrao:
[...]

Issue #135, opened 2020-07-29
Rethinking runCohortDiagnostics function interface


ablack3:
I am very happy about being able to use the CohortDiagnostics package and really appreciate what has been created. It is a wonderful tool. I'm just throwing out an idea to consider here based on some ideas I learned at a workshop on function design. (Relevant slides are [here](https://drive.google.com/file/d/1L-sWi95KLJA4oJ30xLOlEozL0UNlsHjO/view?usp=sharing) for anyone interested)

Perhaps the runCohortDiagnostics() interface could be improved. As a user of this function my primary questions are 

1. What data do I need to supply to the function?
2. What external dependencies need to be met to run the function?
3. What will this function return or what action will this function perform? (i.e. How will the world be different after running the function?)

The function interface is
```
runCohortDiagnostics(
  packageName = NULL,
  cohortToCreateFile = "settings/CohortsToCreate.csv",
  baseUrl = NULL,
  cohortSetReference = NULL,
  connectionDetails = NULL,
  connection = NULL,
  cdmDatabaseSchema,
  oracleTempSchema = NULL,
  cohortDatabaseSchema,
  cohortTable = "cohort",
  cohortIds = NULL,
  inclusionStatisticsFolder = NULL,
  exportFolder,
  databaseId,
  databaseName = databaseId,
  databaseDescription = "",
  runInclusionStatistics = TRUE,
  runIncludedSourceConcepts = TRUE,
  runOrphanConcepts = TRUE,
  runTimeDistributions = TRUE,
  runBreakdownIndexEvents = TRUE,
  runIncidenceRate = TRUE,
  runCohortOverlap = TRUE,
  runCohortCharacterization = TRUE,
  covariateSettings = FeatureExtraction::createDefaultCovariateSettings(),
  minCellCount = 5,
  incremental = FALSE,
  incrementalFolder = exportFolder
)
```
The arguments can be divided into data arguments and detail arguments. Data arguments provide the function with the data it needs to run and detail arguments modify the function's behavior.

**What do I need to supply this function?**

- Either packageName or baseUrl  (a way to obtain the cohort json definitions)
- Either connection or connectionDetails (a way to connect to a CDM)
- Either cohortToCreateFile or cohortSetReference (The cohorts to produce diagnostics for. Which one is required depends on whether packageName or baseUrl is provided.)
- Where to save the output (exportFolder)
- Location of inclusion statistics folder (This is an input to the function which will be read in from a file)
[...]

PR #685, opened 2021-11-19
Refactor run function to improve usability and clean up usage documentation

### LINKED ISSUE #682: Tidy up documentation for usage
There are currently 4 vignettes listed for using the package - I propose replacing these with a single usage document.

Note that many of these use WebApi, we should try and remove this dependency (#677) and simplify the calling interface (#135) so can be completed as part of this step. In addition to having these docs written a valuable contribution would be for a user to check the documentation to ensure it's clarity.

azimov:
Resolves #682, #135 and #677

The objective of this PR is to simplify the input functions to making running this less complicated.

* Created a new function `executeDiagnostics` which aims to (eventually) replace `runCohortDiagnostics` as the main entry point for the package.

- This Function does not take package or reference to ROhdsiWebAPi as an input. It just takes a data.frame. This goes along with the function `loadCohortsFromPackage` which does what the runCohortDiagnostics function does but in a separate step.

* `InstantiateCohortSet` function was also modified to allow this cohort data.frame as a parameter
* placed all of the diagnostic components in individual `execute*Diagnostics` functions for no reason other than my own sanity trying to unravel a 1200 line function call. (note that these functions are not intended to be called externally but there could be improvements in the way they are called, named and parameterized) 
* Updated the vignettes for running cohort diagnostics
- Running cohort diagnostics now describes how to run cohort diagnostics (and not just create cohorts)
- Removed the various modalities as this was a significant source of confusion for users
- The packageMode approach is now emphasised as the preferred method, though a section on how to create cohort inputs with `ROhdsiWebApi` is included.
- The guide is designed to imply (but not explicitly state) that power users can use as inputs the `cohort` parameter as long as they conform to a `data.frame` with `cohortId,  cohortName, logicDescription, json` and `sql` field headers

* Updated usage of tempEmulationSchema to be consitent with other HADES packages and use `getOption`

The `runCohortDiagnostics` entry point has only changed in a few minor ways that will not break functionality or backwards compatibility. However, warnings are added to note that this will change in a future version. The objective is for new users of the package to change to the `executeDiagnostics` API before removing the old settings that load cohorts directly.

I also note that there is usage of `inclusionStatisticsFolder` as a parameter. I think this should be phased out as it is less clean for users than having them use tables in the database. The convention, however, appears to be to use temp tables though so this might be a complicated change.


R/CohortConstruction.R
```diff
@@ -78,76 +78,92 @@ makeBackwardsCompatible <- function(cohorts) {
...
+  cohorts$sql <- sapply(cohorts$name, getSql)
+
+  getJson <- function(name) {
+    pathToJson <-
+      system.file("cohorts", paste0(name, ".json"), package = packageName)
+    checkmate::assertFile(
+      x = pathToJson,
       access = "r",
-      extension = "csv",
+      extension = ".sql",
```
schuemie, line 156: I don't think JSON files have a .sql extension?

schuemie (dismissed):

schuemie:
Excellent work! No major comments from me.

Regarding the `inclusionStatisticsFolder`: As a general rule (in R or any other programming language), functions should not have unintended side-effects. With respect to OHSDI code, this is especially relevant when it comes to the database environment: we should not be created non-temporary tables that we did not tell the user about. So either we allow the user to specify / alter the names of all the tables we're going to create, or we use temp tables instead. In the case of all the statistics tables needed by Circe cohorts (I think we're currently at 5 and counting? A new one was added recently), I think asking the user to check the names of all these tables can be confusing, especially when we can use temp tables instead and just store the data locally. I was hoping we could limit the number of permanent tables to 1 (the cohort table). What do you think?

Pulling in @anthonysena , as this is also relevant for the upcoming [CohortGenerator](https://github.com/OHDSI/CohortGenerator) package.

chrisknoll:
Using temporary tables means that we're left with this awkwardness of having to store some of the results of generation in a permanent table (cohort) and others in a filesystem. 

I don't want to steal @anthonysena thunder, but the cohort generation package is going to have a series of tables included with it to store generation statistics.  Although this will come primarily from Circe-based definitions, there's no issue if a custom cohort generation script wants to write their own statistics to those tables for purposes of diagnostics, since these statistics tables are going to be part of the 'standard cohort generation schema'.

And, I hate to have to clear the record here, but:  the tables involved with cohort generation was stable from 2015-2019, with only [1 table added in 2019](https://github.com/OHDSI/circe-be/commit/98911c6e19a8b013fd433d28acd885b9e83b3ae5), and there are no plans to add more.  I wouldn't call Jan 2019 'recent' (will be 3 years ago in January 2022), so I don't think we should worry about tables being added at random or with any frequency.

schuemie:
Well, I could make a argument for the distinction: although both types of tables are the results of cohort generation, the cohort table itself is the one thing that needs to be in the database, because we're going to be joining it will all sorts of tables in the CDM later on. The statistics tables are exactly that, statistics about what happened during the cohort generation process. Once generated, I see no reason for them to remain on the server. We're not going to join them with anything in the CDM.

In general, we should generate as few permanent tables as possible, to avoid creating a huge mess in the user's schema. If there's no reason for keeping a table (like there is for study-package-generated-cohort statistics), we shouldn't keep them.

anthonysena:
I'm meeting with @chrisknoll and @azimov later to go through this in detail and we'll review the overall workflow for using Cohort Diagnostics with these points in mind.

ablack3:
This is great progress @azimov! Thank you for working on this!

I tried creating a minimal but still realistic example using the `executeDiagnostics` function (pasted below). 

- I got an error about webApiCohortId being missing from the cohort table. Is this a required field for the cohort table? 
- Since the cohortGenerator package and cohortDiagnostics both use a dataframe with cohorts it seems like a good idea to standardize that structure between the two packages. Currently they are slightly different. 
- Is the only reason I need to pass the cohort json and sql in to executeDiagnostics is so that it can be included in the cohort diagnostics app? Is there any guarantee that the json, sql, and cohort table which I pass into executeDiagnostics actually refer to the same cohort definition or is it up to the user to make sure that the cohort table, sql, and json are all coming from the same cohort definition? I'm thinking about the possibility that users may accidentally create cohortDiagnostics on output that does not match the cohort definition json.



```
library(CohortDiagnostics)
library(dplyr)

# First construct a data frame with the cohorts to generate
cohortsToCreate <- CohortGenerator::createEmptyCohortSet()
cohortJsonFiles <- list.files(path = here::here("work/json"), full.names = TRUE)
i=1
# for (i in 1:length(cohortJsonFiles)) {
  cohortJsonFileName <- cohortJsonFiles[i]
  cohortFullName <- tools::file_path_sans_ext(basename(cohortJsonFileName))
  cohortJson <- CohortGenerator::readCirceExpressionJsonFile(cohortJsonFileName)
  cohortJson <- CohortGenerator::readCirceExpressionJsonFile(cohortJsonFileName)
  cohortExpression <- CohortGenerator::createCirceExpressionFromFile(cohortJsonFileName)
  cohortsToCreate <- rbind(cohortsToCreate, data.frame(cohortId = i,
                                                       cohortFullName = cohortFullName, 
                                                       sql = CirceR::buildCohortQuery(cohortExpression, 
                                                                                      options = CirceR::createGenerateOptions(generateStats = FALSE)),
                                                       json = cohortJson,
                                                       stringsAsFactors = FALSE))
# }

cohortsToCreate <- tibble(cohortsToCreate)
cohortsToCreate 

outputFolder <- "~/temp"

cd <- createConnectionDetails("postgresql", user = "postgres", password = "", server = "localhost/covid")

cohortsGenerated <- CohortGenerator::generateCohortSet(connectionDetails = cd,
                                                       cdmDatabaseSchema = "cdm5",
                                                       cohortDatabaseSchema = "cdm5",
                                                       cohortTable = "temp_cohort",
                                                       cohortSet = cohortsToCreate,
                                                       createCohortTable = TRUE,
                                                       incremental = FALSE,
                                                       incrementalFolder = file.path(outputFolder, "RecordKeeping"),
                                                       inclusionStatisticsFolder = outputFolder)


con <- connect(cd)


dbGetQuery(con, "select * from cdm5.temp_cohort") %>% 
  count(cohort_definition_id)


# convert the cohort dataframe to the format required by cohortDiagnostics
cohorts <- cohortsToCreate %>% 
  mutate(logicDescription = cohortFullName) %>%
  rename(cohortName = cohortFullName) %>% 
  select(cohortId, cohortName, logicDescription, sql, json)
  

executeDiagnostics(cohorts = cohorts,
                   exportFolder = "work/export",
                   cohortTable = "tmp_cohort",
                   cohortDatabaseSchema = "cdm5",
                   cdmDatabaseSchema = "cdm5",
                   databaseId = "synthea-covid",
                   connectionDetails = cd)

```

**Error in runCohortDiagnostics(cohorts = cohorts, exportFolder = exportFolder, : 
The following required fields not found in cohort table: webApiCohortId**

R/RunDiagnostics.R
```diff
@@ -76,13 +77,14 @@
 #' @export
 runCohortDiagnostics <- function(packageName = NULL,
                                  cohortToCreateFile = "settings/CohortsToCreate.csv",
+                                 cohorts = NULL,
```
anthonysena, line 80: Could we rename this parameter to `cohortSet` instead?

R/RunDiagnostics.R
```diff
@@ -1203,26 +570,151 @@ runCohortDiagnostics <- function(packageName = NULL,
...
+#' executeDiagnostics(cohorts = cohorts,
+#'                    exportFolder = "export",
+#'                    cohortTable = "cohort",
+#'                    cohortDatabaseSchema = "results",
+#'                    cdmDatabaseSchema = "cdm",
+#'                    databaseId = "mySpecialCdm",
+#'                    connectionDetails = connectionDetails)
+#' }
+#' @export
+executeDiagnostics <- function(cohorts,
```
anthonysena, line 644: Could we change `cohorts` to `cohortSet` instead?

azimov:
Thanks for testing @ablack3! 

> * I got an error about webApiCohortId being missing from the cohort table. Is this a required field for the cohort table?

This seems to be a problem that occured because I fixed some testing code. The commit I just pushed should fix this problem.

> * Since the cohortGenerator package and cohortDiagnostics both use a dataframe with cohorts it seems like a good idea to standardize that structure between the two packages. Currently they are slightly different.

Yep, @anthonysena is working on this. @chrisknoll will also be working on an export function in ROhdsiWebApi that will be consistent with this input.

> * Is the only reason I need to pass the cohort json and sql in to executeDiagnostics is so that it can be included in the cohort diagnostics app? Is there any guarantee that the json, sql, and cohort table which I pass into executeDiagnostics actually refer to the same cohort definition or is it up to the user to make sure that the cohort table, sql, and json are all coming from the same cohort definition? I'm thinking about the possibility that users may accidentally create cohortDiagnostics on output that does not match the cohort definition json.

This is a fair point. The json and sql are not used (directly) to create cohorts with the new execute definition. They are used, however, to extract concept sets.

For me there is the use-case where some custom SQL has been used to generate a cohort that does not relate to any json definition. I don't want to encourage this as a standard input to the package so we should settle on the `CohortGenerator`package in a future version as the consistent way to generate cohorts. This package will be changed to be consistent with those.

anthonysena:
Sorry @azimov - we settled on `cohortDefinitionSet` instead of `cohortSet` after discussion with @chrisknoll.

azimov:
> Sorry @azimov - we settled on `cohortDefinitionSet` instead of `cohortSet` after discussion with @chrisknoll.

No problem, just saw the email thread and changed it

vignettes/RunningCohortDiagnostics.Rmd
```diff
@@ -126,7 +118,36 @@ The resulting cohort table should include the columns:
...
+                                                                cohortDatabaseSchema = cohortDatabaseSchema,
+                                                                cohortInclusionTable = cohortTableNames$cohortInclusionTable)
+
+CohortGenerator::exportCohortStatsTables(connectionDetails = connectionDetails,
+                                         cohortDatabaseSchema = cdmDatabaseSchema,
+                                         cohortTableNames = cohortTableNames,
+                                         cohortStatisticsFolder = inclusionStatisticsFolder,
+                                         incremental = FALSE)
+
+CohortGenerator::dropCohortStatsTables(connectionDetails = connectionDetails,
```
chrisknoll, line 141: I'm wondering about adding the dropChortStatsTable() call in this example.  It's good to have a cleanup function but, is the expected workflow is to generate -> export -> drop tables?  I was expecting more of a multi-generate and when you're done with your study you can drop the tables.

azimov: This is reasonable, it isn't strictly necessary for this guide. I'll remove it

chrisknoll (approved):

vignettes/RunningCohortDiagnostics.Rmd
```diff
@@ -21,53 +21,188 @@ knitr::opts_chunk$set(
...
+                                       cohortDatabaseSchema = cohortDatabaseSchema,
+                                       cohortTableNames = cohortTableNames)
+
+CohortGenerator::dropCohortStatsTables(connectionDetails = connectionDetails,
+                                       cohortDatabaseSchema = cohortDatabaseSchema,
+                                       cohortTableNames = cohortTableNames)
```
anthonysena, line 161: This appears to be a duplicate code block from above.

anthonysena (changes requested):
Just some feedback on things I saw when reviewing the vignette

anthonysena (approved):

PR #706, opened 2021-12-16
Tests on different database platforms

### LINKED ISSUE #704: Add tests for different database platforms
There are currently no automated tests whether CohortDiagnostics runs on the different database platforms. These should be added, using the database platforms in the OHDSI CI testing environment (SQL Server, PostgreSQL, Oracle, RedShift, Bricks). We might be able to add BigQuery as well. 

These tests would need to be fairly light, so for example the number of cohorts need to be kept low (I think everything can be tested with 2 cohorts)

azimov:
Resolves #704


tests/testthat/test-againstCdm.R
```diff
@@ -109,18 +100,19 @@ test_that("Cohort diagnostics in incremental mode", {
...
   testthat::expect_lt(secondTime[1], firstTime[1])
-  
+
   # generate premerged file
   preMergeDiagnosticsFiles(dataFolder = file.path(folder, "export"))
   testthat::expect_true(file.exists(file.path(folder, "export", "PreMerged.RData")))
-  
+
   output <- read.csv(file.path(folder, "export", "covariate_value.csv"))
   expect_equal(output$sum_value[2], -minCellCountValue)
```
azimov, line 115: These last 2 tests fail on the non-eunomia platforms but I'm not sure what they're actually trying to test.

schuemie: These are specific to the exact data in Eunomia and should be removed. I think we're good with the remaining tests.

.github/workflows/R_CMD_check_Hades.yml
```diff
@@ -44,6 +44,11 @@ jobs:
       CDM5_SQL_SERVER_PASSWORD: ${{ secrets.CDM5_SQL_SERVER_PASSWORD }}
       CDM5_SQL_SERVER_SERVER: ${{ secrets.CDM5_SQL_SERVER_SERVER }}
       CDM5_SQL_SERVER_USER: ${{ secrets.CDM5_SQL_SERVER_USER }}
+      #CDM5_REDSHIFT_CDM_SCHEMA: ${{ secrets.CDM5_REDSHIFT_CDM_SCHEMA }}
+      #CDM5_REDSHIFT_OHDSI_SCHEMA: ${{ secrets.CDM5_REDSHIFT_OHDSI_SCHEMA }}
```
azimov, line 48: @schuemie  please note that this schema setting fails with these tests. The tests are designed to skip if these settings aren't present, when fixed the build for redshift can work by uncommenting these lines.

azimov:
In my opinion this is ready to be merged. These tests work but there are two considerations:
* The redshift test environment doesn't support constructing tables; which is required by CohortGenerator
* The code coverage drops slightly. Given that we're now testing on more platforms I don't feel that should be a show stopper.

anthonysena:
> The redshift test environment doesn't support constructing tables; which is required by CohortGenerator

Tagging @leeevans for input - perhaps this requires the use of a different schema. We can discuss offline.

R/CohortComparisonDiagnostics.R
```diff
@@ -69,7 +69,7 @@ computeCohortOverlap <- function(connectionDetails = NULL,
   
   sql <- SqlRender::loadRenderTranslateSql(
     "CohortOverlap.sql",
-    packageName = "CohortDiagnostics",
+    packageName= utils::packageName(),
```
schuemie, line 72: This somehow seems less safe, because we're no long explicit about which package contains the resource. What is the reason for this change?

tests/testthat/helper.R
```diff
@@ -0,0 +1,19 @@
+#' utility function to make sure connection is closed after usage
+with_dbc_connection <- function(connection, code) {
+  on.exit({
+    DatabaseConnector::disconnect(connection)
+  })
+  eval(substitute(code), envir = connection, enclos = parent.frame())
+}
+
+#' Only works with postgres > 9.4
+.pgTableExists <- function(connection, schema, tableName) {
```
schuemie, line 10: Couldn't we use [`DatabaseConnector::getTableNames()`](http://ohdsi.github.io/DatabaseConnector/reference/getTableNames.html) here to check if the table exists?

tests/testthat/setup.R
```diff
@@ -0,0 +1,117 @@
...
+
+dbms <- getOption("dbms", default = "sqlite")
+message("************* Testing on ", dbms, " *************")
+
+if (dir.exists(Sys.getenv("DATABASECONNECTOR_JAR_FOLDER"))) {
+  jdbcDriverFolder <- Sys.getenv("DATABASECONNECTOR_JAR_FOLDER")
+} else {
+  jdbcDriverFolder <- tempfile("jdbcDrivers")
+  dir.create(jdbcDriverFolder, showWarnings = FALSE)
+  DatabaseConnector::downloadJdbcDrivers("postgresql", pathToDriver = jdbcDriverFolder)
```
schuemie, line 13: Why is PostgreSQL special?

tests/testthat/helper.R
azimov: I think this is actually some legacy hack where there was a problem calling `getTableNames` between different database connections (the tables are created with a different connection, probably to increase test coverage), I'll try to remove this.

R/CohortComparisonDiagnostics.R
azimov: Trusting the R namespacing is cleaner to me; it also prevents copy paste errors when people make study packages. I would even prefer a usage of a constant to propagating the reuse of the same string throughout projects.

tests/testthat/helper.R
azimov: Update: use of getTableNames failed in a previous commit, I'm not sure why but it's possibly related to the information schema or some cache being refreshed slower than these tests require. For this reason I reverted the change.

chrisknoll (approved):

Issue #707, opened 2021-12-17
Remove cohort construction code


schuemie:
If we add a dependency on the new [CohortGenerator](https://ohdsi.github.io/CohortGenerator/) package, we can remove all code for generating cohorts from this package.

azimov:
I started a branch for this but there are some issues to resolve. Once these changes are made this can be added to release 2.2, which already references CohortGenerator in the vignettes.

For now, I was going to keep `instantiateCohorts` and `createCohortTable` but include a deprecation warning. This will give any existing users time to change code in their study packages.

gowthamrao:
@azimov i know issuing deprecation warning is best practice - but if its cleaner and easier, i am thinking it would be better to remove the code from Cohort Diagnostics.

Studies have been using renv for sometime now - so i feel like this wont disrupt much.

azimov:
This can be removed before the 2.2 release - the new branch #720 removes the code but it will break any existing packages. If we do this now it might be desirable to only allow the new `executeDiagnostics` function to be used and remove `runCohortDiagnostics` as this is designed to only work with a CohortDefinitionSet in the form expected by CohortGenerator

PR #708, opened 2021-12-17
add Cohort generator dependency

azimov:
Draft PR as it includes changes from #706.

Starts to resolve Issue #707 by deprecating instantiate cohort functions.


R/CohortConstruction.R
```diff
@@ -458,12 +461,31 @@ getInclusionStats <- function(exportFolder,
     ))
   }
   if (nrow(subset) > 0) {
+    if (is.null(inclusionStatisticsFolder)) {
```
azimov, line 464: @anthonysena I'd appreciate it if you approve of this method for extracting the cohort stats data; it outputs the csv files to a temp folder then imports them in to cohort diagnostics.

This is a bit messy but I think its ok for now because it minimises changes in the cohort diagnostics code base. However, I think a future version of CohortGenerator we should include functionality to export the CohortStats to data frames directly.

R/CohortComparisonDiagnostics.R
```diff
@@ -69,7 +69,7 @@ computeCohortOverlap <- function(connectionDetails = NULL,
   
   sql <- SqlRender::loadRenderTranslateSql(
     "CohortOverlap.sql",
-    packageName = "CohortDiagnostics",
+    packageName= utils::packageName(),
```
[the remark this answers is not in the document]
azimov: when `packageName()` is called it uses the namespace/environment of the calling function. In these cases, they are always within the `CohortDiagnostics` namespace. Using the string is a potential source of error that we should try and avoid (especially in study packages which are likely cloned from skeletons).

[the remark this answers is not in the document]
gowthamrao: ok - sounds good!

R/CohortConstruction.R
anthonysena: Hey @azimov this looks good from my review. Just a question that I had after reading through the code - CohortGenerator is invoked under the conditions that there is a non-zero set of cohorts and a null `inclusionStatisticsFolder`. Since this appears to be a private method, is there a condition where the `inclusionStatisticsFolder` would be non-null and skip the retrieval? 

On your other point: I agree we can add a method to CohortGenerator that returns the cohort stats as a data frame so you don't have to do this two-step on the file system.

azimov: Thanks @anthonysena yes, there is the case where `inclusionStatisticsFolder` can be set by the user; this would be the old method of calling `runCohortDiagnostics` which will be retired in a future version but is maintained here for backwards compatibility.

R/CohortConstruction.R
```diff
@@ -558,6 +579,7 @@ instantiateCohortSet <- function(connectionDetails = NULL,
                                  createCohortTable = TRUE,
                                  incremental = FALSE,
                                  incrementalFolder = NULL) {
+  warning("This function will be removed in a future version. Please see https://github.com/ohdsi/CohortGenerator for future generation of cohorts")
```
gowthamrao, line 582: can this just be a note, instead of warning (trying to avoid a red line in log)

R/RunDiagnostics.R
```diff
@@ -88,8 +91,9 @@ runCohortDiagnostics <- function(packageName = NULL,
                                  cohortDatabaseSchema,
                                  vocabularyDatabaseSchema = cdmDatabaseSchema,
                                  cohortTable = "cohort",
+                                 cohortTableNames = CohortGenerator::getCohortTableNames(cohortTable = cohortTable),
```
gowthamrao, line 94: i suggest we add some assertChecks to see if the input values conform to expected

R/RunDiagnostics.R
```diff
@@ -143,6 +149,8 @@ runCohortDiagnostics <- function(packageName = NULL,
     databaseDescription <- databaseId
   }
   
+  cohortTable <- cohortTableNames$cohortTable
```
gowthamrao, line 152: if we dont do assert check, what if this is not a named list object, or cohortTableNames$cohortTable returns NULL

R/RunDiagnostics.R
```diff
@@ -647,18 +660,22 @@ writeResultsZip <- function(exportFolder, databaseId, vocabularyVersion, vocabul
...
                                exportFolder,
                                databaseId,
                                connectionDetails = NULL,
+                               connection = NULL,
                                cdmDatabaseSchema,
                                tempEmulationSchema = getOption("sqlRenderTempEmulationSchema"),
                                cohortDatabaseSchema,
-                               vocabularyDatabaseSchema = cdmDatabaseSchema,
                                cohortTable = "cohort",
+                               cohortTableNames = CohortGenerator::getCohortTableNames(cohortTable = cohortTable),
```
gowthamrao, line 675: recommend adding assertChecks

gowthamrao (approved):
@azimov  i would ask that you consider using the assertCheck that has been used in other places in the package to ensure the input to exposed functions are conformant to expected. If not, provide an informative error message and write to log

 errorMessage <- checkmate::makeAssertCollection()

anthonysena (approved):
Looks good from my review.

PR #720, opened 2022-01-19
Remove and refactor code that is replaced by CohortGenerator package

### LINKED ISSUE #707: Remove cohort construction code

azimov:
Resolves #707


R/RunDiagnostics.R
```diff
@@ -75,54 +71,85 @@
...
+                               tempEmulationSchema = getOption("sqlRenderTempEmulationSchema"),
+                               cohortDatabaseSchema,
+                               cohortTable = "cohort",
+                               cohortTableNames = CohortGenerator::getCohortTableNames(cohortTable = cohortTable),
+                               vocabularyDatabaseSchema = cdmDatabaseSchema,
+                               cohortIds = NULL,
+                               databaseName = databaseId,
+                               databaseDescription = databaseId,
+                               cdmVersion = 5,
+                               runInclusionStatistics = TRUE,
```
gowthamrao, line 128: could you reorder the parameter to be as identical as possible to the original runCohortDiagnostics

e.g. databaseId and databaseDescription are now separated 

i think it would make it easier for the transition

R/RunDiagnostics.R
```diff
@@ -147,28 +174,12 @@ runCohortDiagnostics <- function(packageName = NULL,
     takepackageDependencySnapshot() %>%
     RJSONIO::toJSON(digits = 23, pretty = TRUE)
 
-
   exportFolder <- normalizePath(exportFolder, mustWork = FALSE)
   incrementalFolder <- normalizePath(incrementalFolder, mustWork = FALSE)
 
-  if (!is.null(inclusionStatisticsFolder)) {
```
gowthamrao, line 154: by removing inclusionStatisticsFolder - we are asserting that
- inclusion rules table always have to be on remote db and should not have been dropped prior to starting cohort diagnostics
- i think we should make this explicit and clear in some documentation

R/RunDiagnostics.R
```diff
@@ -147,28 +174,12 @@ runCohortDiagnostics <- function(packageName = NULL,
...
-  
   start <- Sys.time()
   ParallelLogger::logInfo("Run Cohort Diagnostics started at ", start)
-  
-  if (!is.null(oracleTempSchema) & is.null(tempEmulationSchema)) {
-    tempEmulationSchema <- oracleTempSchema
-    warning('OracleTempSchema has been deprecated by DatabaseConnector')
-  }
-  
-  databaseId <- as.character(databaseId)
```
gowthamrao, line 171: The reason why this is forced to be character is because in one network study a databaseId of value integer was provided - that caused problems when readr was guess data types

this maybe less of an issue now - because we are planning to check conformance to results data model

R/RunDiagnostics.R
```diff
@@ -747,7 +745,7 @@ runCohortDiagnostics <- function(packageName = NULL,
   )
 
   # Add all to zip file -------------------------------------------------------------------------------
-  writeResultsZip(exportFolder, databaseId, vocabularyVersion, vocabularyVersionCdm)
+  writeResultsZip(exportFolder, databaseId, vocabularyVersion, vocabularyVersion)
```
gowthamrao, line 748: why was vocabularyVersionCdm changed to vocabularyVersion ?

gowthamrao (changes requested):

R/RunDiagnostics.R
azimov: The user doesn't need to do anything at all; CohortDiagnostics now uses CohortGenerator to populate and export the results. I don't know what the document should say? Don't delete the cohort tables until you've finished the analysis? This seems somewhat obvious.

R/RunDiagnostics.R
azimov: Both these last two parameters are actually no longer used since the recent meta-data patch. I will remove them.

R/RunDiagnostics.R
azimov: The parameters have actually changed quite significantly so this isn't really possible; this is one reason for the change of name - to highlight that there is a big difference. I have moved database name and description to be next to the databaseId but Its best practice to keep required parameters first and ones with default values after it. With functions like this users should also try and use all named parameters because there are so many.

gowthamrao: I see; i am fine with either way you want to take it.

R/RunDiagnostics.R
gowthamrao: i think u r right

and if they have dropped the inclusion rules table from the server - they would not get inclusion rule statistics diagnostics.

gowthamrao: In terms of documentation -- if some folks dont see inclusion rule diagnostics, when one was expected - the obvious first question should be - did u drop the inclusion rule tables before running cohort diagnostics

R/RunDiagnostics.R
```diff
@@ -745,15 +747,15 @@ executeDiagnostics <- function(cohortDefinitionSet,
...
+  writeResultsZip(exportFolder, databaseId)
   ParallelLogger::logInfo("Computing all diagnostics took ",
                           signif(delta, 3),
                           " ",
                           attr(delta, "units"))
 }
 
 
-writeResultsZip <- function(exportFolder, databaseId, vocabularyVersion, vocabularyVersionCdm) {
+writeResultsZip <- function(exportFolder, databaseId) {
```
gowthamrao, line 758:
```
writeResultsZip <- function(exportFolder, databaseId, vocabularyVersion, vocabularyVersionCdm) {
  ParallelLogger::logInfo("Adding results to zip file")
```
thanks for making the change

gowthamrao (approved):

Issue #539, opened 2021-09-23
Replace pre-merged file with Sqlite DB/Andromeda file


azimov:
After discussions with @anthonysena, we noted that an implementation caveat with CohortDiagnostics and LSC is that there is a use of both a postgres database implementation and use of PreMerged files for use in the shiny apps. Ideally, we would like a single data storage model that works whether the data is big or small.

Is there a good reason why the study organiser can't convert the CSV files to an Sqlite database? This would unify the data model which would aid in developing more modular applications in the future. We could also resolve the postgres only functionality that is included in the Shiny apps to support hosting on any DatabaseConnector/SqlRender compatible platforms (though there may be some caveats for storage of json strings/cohort definitions in platforms like redshift).

gowthamrao:
I think @leeevans is interested in this idea https://github.com/OHDSI/CohortDiagnostics/issues/317

We dont use sqlite because its not designed for concurrent queries. We want the db that hosting the result set to be able to support multiple connections - including SQL based queries from users. Shiny app is not the only user of the connection

gowthamrao:
@azimov as long as we can support the SQL translations, we can achieve this

gowthamrao:
If we decide to pursue this, i recommend we work on it - after next release

leeevans:
@azimov I really like this idea. I began working on implementing the premerge file functionality into a standard function in the ohdsisharing package. I paused that activity to work on the Symposium infrastructure. I'm starting to work on that again now. We can create a mergeToSqlite function in the ohdsisharing package to load/merge the csv files into a single sqlite database file.

@gowthamrao made a good point that it is useful to retain the use of csv files at local sites because they can be easily opened and reviewed in simple tools like Excel, so this sqlite file would be created in addition to the csv files generated by the study.

azimov:
@leeevans Agreed, the data storage at sites/from the base analysis should still be in CSV formats so we don't need to touch this.

I see there being a few steps to this:
1) making the sqlite DB from the CSV files in ohdsi sharing
2) Modifying the shiny requests to use only DatabaseConnector calls and taking out any postgres specific queries (if there are any?)
3) Optimising the shiny app load (the premerged file will probably load the app faster)

@gowthamrao this also relates to issue #434 as we will not need to expose the model to the pre-merged file.

gowthamrao:
> making the sqlite DB from the CSV files in ohdsi sharing

[...]

Issue #317, opened 2021-02-03
Enable CohortDiagnostics to run on other DBMSs in addition to Postgres


leeevans:
Cohort Diagnostics is a great tool. It would be very useful to be able to run it on other DBMSs in addition to Postgres. If some advanced Cohort Diagnostics functionality depends on Postgres DBMS specific features, then those features could perhaps be disabled when running on other DBMSs.

azimov:
Cohort diagnostics can run on any OMOP CDM compliant CDM but hosting the Shiny dashboard for large data sets currently only supports postgres. Hosting without postgres is possible with the exported RData objects but this becomes limited with larger data sets. 

Postgres was deliberately selected because its open source and supporting multiple database backends may add additional work. We only expect this use case to occur where people want to share the results of a large study with OHDSI, for example across multiple sites in a network study. 

We would also almost always expect the postgres instance to be separate from any CDMs as the output of Cohort Diagnostics does not contain patient level data.

gowthamrao:
@leeevans is there interest to support non postgres based db for the shiny app. Like @azimov said, this only applies to the post computed result set with no person level data that is used by the shiny app

leeevans:
@azimov thanks for pointing out that the Cohort Diagnostics can be generated (by reading data) from CDMs on any OHDSI (DatabaseConnector) supported DBMS. @gowtham, yes, I guess the scope of this issue is narrowed down to the DBMS support for the cohort diagnostics results database used by the Shiny app.

I see a postgresql dependency in the way the cohort diagnostics results schema tables are created (postgresql only DDL executed) and postgresql database specific upload of the cohort diagnostics results ("PreMerged.Rdata") dataset in the preMergeDiagnosticsFiles() function. There is also a check in the shiny app that stops execution if the database connection DBMS is not 'postgresql'. I know that the scope of this shiny app was to run it only on the OHDSI shiny server postgresql database so this is understandable.

If I could somehow load the 'PreMerged.RData' dataset into a results schema on a Redshift database or a SQL Server database and I removed the postgresql DBMS connection check from the shiny app code would the shiny app work, or are there other additional postgresql dependencies in the shiny app UI/Server code? 

Also, is it possible to use the "PreMerged.RData" dataset with the Shiny app or does it require a (postgresql) database? I agree that using a database for the typically large results data is preferred but I just wondered if executing the shiny app using that pre-merged dataset is an option?

gowthamrao:
[...]

PR #721, opened 2022-01-21
Sqlite shiny backend

azimov:
This is designed for Issue #539 but also (potentially) allows any database platform to act as a backend for the shiny app (see Issue #317).

As sqlite is an RDBMS stored on the disk performance may be slower for smaller data files. However, I expect this will allow users to have much larger results sets than is currently supported by the rdf files (and will likely scale better too).


R/ResultsDataModel.R
```diff
@@ -462,7 +451,8 @@ uploadResults <- function(connectionDetails = NULL,
         } else {
           DatabaseConnector::insertTable(
             connection = connection,
-            tableName = paste(env$schema, env$tableName, sep = "."),
+            tableName = env$tableName,
```
gowthamrao, line 454: i am thinking DatabaseConnector will handle the bulk load based on the system variables? i.e. we dont need to allow an overide in this package?

> 
>   bulkLoad = Sys.getenv("DATABASE_CONNECTOR_BULK_UPLOAD"),
>   useMppBulkLoad = Sys.getenv("USE_MPP_BULK_LOAD"),

R/Shiny.R
```diff
@@ -42,8 +39,7 @@
 #' Launches a Shiny app that allows the user to explore the diagnostics
 #'
 #' @export
-launchDiagnosticsExplorer <- function(dataFolder = "data",
-                                      dataFile = "PreMerged.RData",
+launchDiagnosticsExplorer <- function(sqliteDbPath = "MergedCohortDiagnosticsData.sqlite",
```
gowthamrao, line 42: would this be considered breaking change? support to premerged file is gone?

R/Shiny.R
```diff
@@ -52,23 +48,27 @@ launchDiagnosticsExplorer <- function(dataFolder = "data",
...
+
+  sqliteDbPath <- normalizePath(sqliteDbPath)
+  if (is.null(connectionDetails)) {
+    if (!file.exists(sqliteDbPath)) {
+      stop("Sqlite database", sqliteDbPath, "not found. Please see createMergedSqliteResults")
     }
+
+    resultsDatabaseSchema <- "main"
+    vocabularyDatabaseSchemas <- "main"
+    connectionDetails <- DatabaseConnector::createConnectionDetails(dbms = "sqlite", server = sqliteDbPath)
```
gowthamrao, line 60: does sql lite support the DBPool?

R/Shiny.R
```diff
@@ -110,128 +110,59 @@ launchDiagnosticsExplorer <- function(dataFolder = "data",
     connectionDetails = connectionDetails,
     resultsDatabaseSchema = resultsDatabaseSchema,
     vocabularyDatabaseSchemas = vocabularyDatabaseSchemas,
-    dataFolder = dataFolder,
```
gowthamrao, line 113: this would simplify

R/Shiny.R
```diff
@@ -110,128 +110,59 @@ launchDiagnosticsExplorer <- function(dataFolder = "data",
...
+#' This function combines diagnostics results from one or more databases into a single file. The result is an
+#' sqlite database that can be used as input for the Diagnostics Explorer Shiny app.
 #'
 #' It also checks whether the results conform to the results data model specifications.
 #'
 #' @param dataFolder       folder where the exported zip files for the diagnostics are stored. Use
 #'                         the \code{\link{runCohortDiagnostics}} function to generate these zip files.
 #'                         Zip files containing results from multiple databases may be placed in the same
 #'                         folder.
-#' @param tempFolder       A folder on the local file system where the zip files are extracted to. Will be cleaned
```
gowthamrao, line 134: maybe we couldc, if tempFolder parameter is being used

- put a stop error (not warning)
- but before stopping - check if there is preMergedFile in the tempFolder
- state that premerged file support was stopped in version x and after, and replace with MergedCohortDiagnosticsData.sqlite.
- show user how to create MergedCohortDiagnosticsData.sqlite

R/Shiny.R
```diff
@@ -110,128 +110,59 @@ launchDiagnosticsExplorer <- function(dataFolder = "data",
...
-#'                         up when the function is finished. Can be used to specify a temp folder on a drive that
-#'                         has sufficient space if the default system temp space is too limited.
-#'
+#' @param sqliteDbPath     Output path where sqlite database is placed
+#' @param overwrite        (Optional) overwrite existing sqlite lite db if it exists.
 #' @export
-preMergeDiagnosticsFiles <-
```
gowthamrao, line 139: i am not sure what the best practice is -- if someone has an old script with preMergeDiagnosticsFiles, would they just get a function not found type of error? or should we keep a preMergeDiagnosticsFiles function -- all it would do is give an informative stop error and recommend the use of createMergedResultsFile function

R/Shiny.R
```diff
@@ -110,128 +110,59 @@ launchDiagnosticsExplorer <- function(dataFolder = "data",
...
-    
-    # Storing output in an environment for now. If things get too big, we may want to write
-    # directly to CSV files for insertion into database:
-    newEnvironment <- new.env()
-    
-    processTable <- function(tableName, env) {
-      ParallelLogger::logInfo("Processing table ", tableName)
-      csvFileName <- paste0(tableName, ".csv")
-      data <- dplyr::tibble()
-      for (i in 1:nrow(zipFiles)) {
```
gowthamrao, line 177: @azimov - wow, all these checks will be gone because sqlite will enforce column names, data types and duplicate rows!

inst/shiny/DiagnosticsExplorer/global.R
```diff
@@ -1,95 +1,80 @@
...
+    dbms = dbms,
+    server = Sys.getenv("shinydbServer"),
+    port = Sys.getenv("shinydbPort", unset = 5432),
+    user = Sys.getenv("shinydbUser"),
+    password = Sys.getenv("shinydbPw")
+  )
+
+  resultsDatabaseSchema <- Sys.getenv("shinydbResultsSchema", unset = "thrombosisthrombocytopenia")
+  vocabularyDatabaseSchemas <- resultsDatabaseSchema
+  alternateVocabularySchema <-  Sys.getenv("shinydbVocabularySchema", unset = c("vocabulary"))
```
gowthamrao, line 46: alternateVocabularySchema or referenceVocabulary allowed to compare the impact of vocabulary changes in diagnostics. There can be more than one, and the user gets to toggle between them in the shiny application.

i dont remember if there was support for the the referenceVocabulary in the function that creates shinySettings.

inst/shiny/DiagnosticsExplorer/global.R
```diff
@@ -1,95 +1,80 @@
...
+      drv = DatabaseConnector::DatabaseConnectorDriver(),
+      dbms = connectionDetails$dbms,
+      server = connectionDetails$server(),
+      port = connectionDetails$port(),
+      user = connectionDetails$user(),
+      password = connectionDetails$password(),
+      connectionString = connectionDetails$connectionString()
+    )
 } else {
-  writeLines("Using settings provided by user")
```
gowthamrao, line 59: nice - simplification

inst/settings/resultsDataModelSpecification.csv
```diff
@@ -81,12 +81,12 @@ covariate_value_dist,p_25_value,float,Yes,No,No,Yes
 covariate_value_dist,p_75_value,float,Yes,No,No,Yes
 covariate_value_dist,p_90_value,float,Yes,No,No,Yes
 covariate_value_dist,database_id,varchar,Yes,Yes,No,Yes
-database,database_id,varchar,Yes,Yes,No,Yes
-database,database_name,varchar,No,No,No,Yes
-database,description,varchar,No,No,No,Yes
-database,is_meta_analysis,varchar(1),Yes,No,No,Yes
-database, vocabulary_version, varchar, No, No, Yes, Yes
-database, vocabulary_version_cdm, varchar, No, No, Yes, Yes
+database_info,database_id,varchar,Yes,Yes,No,Yes
```
gowthamrao, line 84: this is a breaking change

inst/shiny/DiagnosticsExplorer/R/Results.R
```diff
@@ -564,94 +407,91 @@ getCovariateValueResult <- function(dataSource = .GlobalEnv,
...
 }
 
-resolveMappedConceptSetFromVocabularyDatabaseSchema <- function(dataSource = .GlobalEnv, 
+resolveMappedConceptSetFromVocabularyDatabaseSchema <- function(dataSource,
                                                                 conceptSets,
                                                                 vocabularyDatabaseSchema = 'vocabulary') {
-  if (is(dataSource, "environment")) {
-    stop("Cannot resolve concept sets without a database connection")
+  if (dbms == "sqlite") {
+    stop("Cannot resolve concept sets without a full vocabulary")
```
gowthamrao, line 494: @azimov  this is incorrect. The cohort diagnostics standard output will have vocabulary tables. The concepts in the vocabulary tables will be just enough to resolve the concept set expression (Based on how it was resolved at the time when cohort diagnostics initially ran)

reason why this not supported for premerged file was because we did not convert the sql logic to resolve concept-set-expression to R-dply syntax.

gowthamrao: so now we can use that sql - because the sql is in ohdsi sql and sqlite can use OHDSI sql with translation!

gowthamrao (changes requested):

inst/sql/sql_server/CreateResultsDataModel.sql
```diff
@@ -94,8 +186,8 @@ CREATE TABLE cohort_overlap (
 
 --Table concept
 
-CREATE TABLE concept (
-			concept_id INT NOT NULL,
+CREATE TABLE @results_schema.@concept (
+			concept_id BIGINT NOT NULL,
```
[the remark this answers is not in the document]
azimov: This fixes another bug that was found where concept ids were too large for INT. I can't find reference to it now, will keep looking

R/ResultsDataModel.R
azimov: yes, this is correct DatabaseConnector should handle that. I'm not sure how much difference it makes with sqlite. When we properly test, document and support other platforms this would be useful to include in a vignette.

R/Shiny.R
azimov: Yes, I don't think its a good idea to continue to support the PreMerged.RData file in newer versions of the app. Though it may be beneficial to allow some sort of conversion mechanism. However, I assumed that the zip files are the main results.

R/ResultsDataModel.R
```diff
@@ -236,31 +236,15 @@ createResultsDataModel <- function(connection = NULL,
...
     }
   }
-  schemas <- unlist(
-    DatabaseConnector::querySql(
-      connection,
-      "SELECT schema_name FROM information_schema.schemata;",
-      snakeCaseToCamelCase = TRUE
-    )[, 1]
-  )
-  if (!tolower(schema) %in% tolower(schemas)) {
```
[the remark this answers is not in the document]
azimov: I can add it back but it seemed the check didn't add a lot as the sql execution will stop anyway if the schema doesn't exist.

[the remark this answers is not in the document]
gowthamrao: I see - i am fine with removing, if you think the error message from sql termination would be informative enough

inst/sql/sql_server/CreateResultsDataModel.sql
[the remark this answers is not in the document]
gowthamrao: yea -- i knew there was a reason for the BIGINT, but i wasnt sure if it was a BUG or just a data have conceptId in BIGINT range

R/ResultsDataModel.R
gowthamrao: Well - the major contribution of this PR is the use of SqlRender with translation. i.e. all SQL in DiagnosticExplorer should be in OHDSI Sql form. We are translating the OHDSI SQl to either SQL render or Postgres, but with this framework - we can support any db that is supported by SqlRender, as long as we can have a dbPool connection.

If this logic is true --then BULKLOAD would be useful for the MPP db's that is supported

R/Shiny.R
gowthamrao: Yes - ur assumption is correct. ZipFiles are the main results. So, based on that - i think this breaking change (i.e. not using premerged file) is ok

inst/settings/resultsDataModelSpecification.csv
azimov: Yes, and thinking about it we will want to support importing old zip files that contain "database.csv" files. This needs to be redisgned.

inst/shiny/DiagnosticsExplorer/global.R
azimov: `launchDiagnosticsExplorer` has a `vocabularyDatabaseSchemas` parameter so support for multiple vocabs is possible this way

gowthamrao (approved):

PR #730, opened 2022-02-02
Metadata enhancement

gowthamrao:


inst/sql/sql_server/CreateResultsDataModel.sql
```diff
@@ -362,9 +362,10 @@ CREATE TABLE @results_schema.@index_event_breakdown (
 
 CREATE TABLE @results_schema.@metadata (
 			database_id VARCHAR NOT NULL,
+			start_time VARCHAR NOT NULL,
 			variable_field VARCHAR NOT NULL,
 			value_field VARCHAR,
-			PRIMARY KEY(database_id, variable_field)
+			PRIMARY KEY(database_id, start_time, variable_field)
```
azimov, line 368: will this break backwards compatibility with old runs where metadata.csv doesn't contain a start time?
Could it be made not required and default to something like 0:0:0?

inst/settings/resultsDataModelSpecification.csv
```diff
@@ -121,6 +121,7 @@ index_event_breakdown,database_id,varchar,Yes,Yes,No,Yes
 index_event_breakdown,domain_field,varchar,Yes,Yes,No,Yes
 index_event_breakdown,domain_table,varchar,Yes,Yes,No,Yes
 metadata,database_id,varchar,Yes,Yes,No,Yes
+metadata,start_time,varchar,Yes,Yes,No,Yes
```
azimov, line 124: does isRequired need to be set to true?

gowthamrao: metadata table changes wont break DiagnosticsExplorer because it is not used by DiagnosticsExplorer. Metadata file was mostly explored as a csv file

gowthamrao: You are right, this maybe an optional field

inst/sql/sql_server/CreateResultsDataModel.sql
gowthamrao: I think it can be defaulted to 0:0:0

gowthamrao: Do you know to modify the ddl so that it always has a default value?

gowthamrao: Will it break DiagnosticExplorer - metadata is not used by DiagnosticExplorer right now. its mostly to explore the content as csv

azimov:
```
start_time VARCHAR DEFAULT '0:0:0',
```

Should work

azimov (approved):

Issue #503, opened 2021-07-29
Cohort diagnostic - Translation error for query in  ComputeTimeSeries.sql


Linasulieman:
I am trying to run the CohortDiagnostic function on a database that is saved in a BigQuery schema. When I ran the the runCohortDiagnostics function I got this error that was generated for the query in  ComputeTimeSeries.sql
**The error that I got was:**
Error:
java.sql.SQLException: [Simba]BigQueryJDBCDriver Error executing query job. Message: Column 6 contains an aggregation function, which is not allowed in GROUP BY at [53:19]

SQL:
select cohort_definition_id cohort_id
,period_begin
,calendar_interval
,count() records
,count(distinct subject_id) subjects
,sum(date_diff(if(safe_cast(case 
  		when cohort_end_date >= period_end
  			then period_end
  		else cohort_end_date
  		end  as date) is null,parse_date('%Y%m%d', cast(case 
  		when cohort_end_date >= period_end*
  			then period_end
  		else cohort_end_date
  		end  as string)),safe_cast(case 
  		when cohort_end_date >= period_end
  			then period_end
  		else cohort_end_date
  		end  as date)), if(safe_cast(case 
  		when cohort_start_date >= period_begin
  			then cohort_start_date
  		else period_begin
  		end  as date) is null,parse_date('%Y%m%d', cast(case 
  		when cohort_start_date >= period_begin
  			then cohort_start_date
  		else period_begin
  		end  as string)),safe_cast(case 
  		when cohort_start_date >= period_begin
  			then cohort_start_date
  		else period_begin
  		end  as date)), day) + 1) person_days
,count(case when cohort_start_date >= period_begin
  		and cohort_start_date <= period_end
  		then subject_id
  	else null
  	end) records_incidence
,count(distinct case 
  	when cohort_start_date >= period_begin
  		and cohort_start_date <= period_end
  		then subject_id
  	else null
  	end) subjects_incidence
from test_writable_sulieman.cohorttabletest
inner join ofcjbawzcalendar_periods cp on (
  cohort_start_date >= period_begin
  and cohort_start_date <= period_end
  )
or (cohort_end_date >= period_begin and cohort_end_date <= period_end )
where cohort_definition_id in (101431)
group by 2, 3, 6 ;

**Reason**
[...]

PR #736, opened 2022-02-03
Legacy time series is incorrect and needs to be deprecated

gowthamrao:
It is not used in any studies, sql does not run
https://github.com/OHDSI/CohortDiagnostics/issues/503


R/RunDiagnostics.R
```diff
@@ -123,6 +123,10 @@ runCohortDiagnostics <- function(packageName = NULL,
                                  incremental = FALSE,
                                  incrementalFolder = file.path(exportFolder, "incremental")) {
   
+  if (runTimeSeries) {
```
azimov, line 126: This is incorrect; deprecation means that it will be removed in a future version. Here it's just being removed (which is fine) but the `runTimeSeries` parameter should probably just be removed and we can add this info to the release notes.

NEWS.md
```diff
@@ -1,3 +1,11 @@
+CohortDiagnostics 2.2.2
```
azimov, line 1: Is this intended to be a hotfix for the main branch? Develop includes code for version 3.0.0.

Considering that this is the removal of functionality (though most users probably aren't using it), I suggest it goes in to version 3.

R/RunDiagnostics.R
gowthamrao: i see - so if we incorporate these changes in future release and a user has a legacy code - it would throw an error. 

is that acceptable?

azimov: We're replacing `runCohortDiagnostics` with `executeDiagnostics` anyway; I think we should just remove the parameter; adding it to the release notes will clear up any confusion.

azimov (approved):

PR #748, opened 2022-02-08
Make export data simple

gowthamrao:


R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
...
+  
+  presentInBoth <-
+    intersect(fieldsInDataModel, names(x))
+  presentInDataOnly <-
+    setdiff(names(x), fieldsInDataModel)
+  missingRequiredFields <-
+    setdiff(requiredFieldsInDataModel, presentInBoth)
+  
+  if (length(presentInDataOnly) > 0) {
+    warning(
```
azimov, line 179: I would change this from a warning to a logInfo - our code should just be resilient to users adding optional fields but notifying them is useful

R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
   x[is.null(x)] <- ""
   return(x)
 }
+
+
+
+makeDataExportable <- function(x,
```
azimov, line 121: Can you add some unit tests for this function? This is useful as we want to know that all the checks are properly enforced

azimov (changes requested):

R/CohortConstruction.R
```diff
@@ -145,16 +132,22 @@ getInclusionStatisticsFromFiles <- function(cohortIds = NULL,
         summaryStats = filter(summaryStats, .data$cohortDefinitionId == cohortId),
         simplify = simplify
       )
-    cohortResult$cohortDefinitionId <- cohortId
-    result <- dplyr::bind_rows(result, cohortResult)
+    if (!is.null(cohortResult)) {
```
azimov, line 135: I'm not sure how to add it to this PR but this code is never run in our tests so cohortResult is always null when we run it here.

R/CohortConstruction.R
```diff
@@ -257,45 +250,32 @@ getInclusionStats <- function(connection,
...
-          folder = inclusionStatisticsFolder,
-          simplify = TRUE
-        )
+      getInclusionStatisticsFromFiles(
+        cohortIds = subset$cohortId,
+        folder = inclusionStatisticsFolder,
+        simplify = TRUE
+      )
 
     if (!is.null(stats)) {
```
azimov, line 259: again, it seems stats is always null so we're not testing this behvaiour

R/ConceptIds.R
```diff
@@ -30,59 +30,24 @@ createConceptTable <- function(connection, tempEmulationSchema, cohorts) {
...
-                                       "concept",
-                                       "conceptAncestor",
-                                       "conceptClass",
-                                       "conceptRelationship",
-                                       "conceptSynonym",
-                                       "domain",
-                                       "relationship",
-                                       "vocabulary"
-                                     ),
+                                     vocabularyTableNames = getResultsDataModelSpecifications() %>%
```
azimov, line 39: using this code here is very confusing for a default parameter. I would prefer this to be a function like `getDefaultVocabularyTableNames()`

R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
...
+  
+  resultsDataModel <- getResultsDataModelSpecifications()
+  
+  if (!is.null(databaseId)) {
+    x <- x %>%
+      dplyr::mutate(databaseId = !!databaseId)
+  }
+  
+  fieldsInDataModel <- resultsDataModel %>%
+    dplyr::filter(.data$tableName == !!tableName) %>%
```
azimov, line 138: what is the `!!tableName` for. The double negation implies that it is the tableName so `dplyr::filter(.data$tableName == !!tableName)` is functionally identical.

Tbh I'm not sure why !!tableName doesn't just bomb (!<string>) is clearly undefined; so R's operator precedence is just weird here.

R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
...
+    dplyr::pull(.data$fieldName) %>%
+    snakeCaseToCamelCase() %>%
+    unique()
+  
+  requiredFieldsInDataModel <- resultsDataModel %>%
+    dplyr::filter(.data$tableName == !!tableName) %>%
```
azimov, line 144: Same, remove `!!tableName`

R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
...
+    dplyr::filter(.data$isRequired == "Yes") %>%
+    dplyr::pull(.data$fieldName) %>%
+    snakeCaseToCamelCase() %>%
+    unique()
+  
+  primaryKeyInDataModel <- resultsDataModel %>%
+    dplyr::filter(.data$tableName == !!tableName) %>%
```
azimov, line 151: `!!tableName`

R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
...
+    dplyr::filter(.data$primaryKey == "Yes") %>%
+    dplyr::pull(.data$fieldName) %>%
+    snakeCaseToCamelCase() %>%
+    unique()
+  
+  columnsToApplyMinCellValue <- resultsDataModel %>%
+    dplyr::filter(.data$tableName == !!tableName) %>%
```
azimov, line 158: `!!tableName`

R/Private.R
```diff
@@ -115,3 +115,157 @@ nullToEmpty <- function(x) {
...
+  if (nrow(x) >  nrow(x %>% dplyr::select(dplyr::all_of(primaryKeyInDataModel)) %>% dplyr::distinct())) {
+    stop(
+      " - duplicates found in primary key for table ",
+      tableName,
+      ". The primary keys are: ",
+      paste0(primaryKeyInDataModel, collapse = ", ")
+    )
+  }
+  ## because Andromeda is not handling date consistently -
+  # https://github.com/OHDSI/Andromeda/issues/28
```
azimov, line 207: Is x an andromeda object?

R/RunDiagnostics.R
```diff
@@ -666,7 +671,7 @@ executeDiagnostics <- function(cohortDefinitionSet,
     #16
     "vocabularyVersionCdm",
     #17
-    "vocabularyVersion",
+    # "vocabularyVersion",
```
azimov, line 674: Why is this now commented? Are you no longer storing this in the metadata?

R/RunDiagnostics.R
```diff
@@ -721,7 +726,7 @@ executeDiagnostics <- function(cohortDefinitionSet,
     #16
     as.character(nullToEmpty(cdmSourceInformation$vocabularyVersion)),
     #17
-    as.character(vocabularyVersion),
+    # as.character(vocabularyVersion),
```
azimov, line 729: Again, what's the reason for the comment? Either remove or keep it.

tests/testthat/test-other.R
```diff
@@ -3,3 +3,46 @@ test_that("Check if package is installed", {
   expect_true(CohortDiagnostics:::is_installed('dplyr'))
   expect_false(CohortDiagnostics:::is_installed('abcd'))
 })
+
+
+#check makeDataExportable function
+test_that("Check function makeDataExportable", {
+  cohortCountTableCorrect <- dplyr::tibble(
```
azimov, line 10: great tests!

R/CohortConstruction.R
azimov: Don't worry about it in this PR, I will add an issue at some point to create some unit tests to fully tests this (probably with some mock exported data).

azimov (changes requested):

R/Private.R
gowthamrao: bang bang!! u no like it?

R/ConceptIds.R
gowthamrao: ok - i will add a function

azimov (approved):

Issue #738, opened 2022-02-03
Cohort relationship logic


gowthamrao:

gowthamrao:
Probably later in 3

PR #758, opened 2022-02-10
Cohort Relationship logic

gowthamrao:
https://github.com/OHDSI/CohortDiagnostics/issues/738


azimov (changes requested):
So the code looks good but I would really like to see tests of the cohort relationship function.

I think this should take the form of a few man made cohorts that have specific dates with time windows. This would make the code a bit easier to understand (especially as there is no visualization available yet).

By this I mean inserting cohort tables directly with dates that overlap in specific ways at specific time windows so you will be able to test that cohorts have relationships when you think they should or shouldn't etc.

The tests should run against the cohort relationships function directly and not the executeDiagnostics function.

tests/testthat/test-moduleCohortRelationship.R
```diff
@@ -0,0 +1,52 @@
+test_that("Testing cohort relationship logic", {
+  # manually create cohort table and load to table
+  
+  # connection <-
```
azimov, line 4: I think you want to create a connection with the `connectionDetails` object created in `setup.R`.

I've also been using `with_dbc_connection` in the helpers.R utilities to ensure that the connection closes.

For example

```{r}
connection <- DatabaseConnector::connect(connectionDetails)
with_dbc_connection(connection, {
   ...

  runCohortRelationshipDiagnostics(
      connection = connection,
      cohortDatabaseSchema = NULL,
      cdmDatabaseSchema = cdmDatabaseSchema,
      cohortTable = "#cohort_relationship",
      targetCohortIds = c(1, 2, 3),
      comparatorCohortIds = c(10, 11, 12),
      relationshipDays = dplyr::tibble(startDay = temporalStartDays,
                                       endDay = temporalEndDays)
  )

  ...
}
```

tests/testthat/test-moduleCohortRelationship.R
```diff
@@ -0,0 +1,52 @@
...
+    connection = connection,
+    cohortDatabaseSchema = NULL,
+    cdmDatabaseSchema = cdmDatabaseSchema,
+    cohortTable = "#cohort_relationship",
+    targetCohortIds = c(1, 2, 3),
+    comparatorCohortIds = c(10, 11, 12),
+    relationshipDays = dplyr::tibble(startDay = temporalStartDays,
+                                     endDay = temporalEndDays)
+  )
+
```
azimov, line 50: it would be good to see what the expected results are for the cohorts

tests/testthat/test-moduleCohortRelationship.R
```diff
@@ -0,0 +1,52 @@
...
+  
+  comparatorCohort <- dplyr::tibble(cohortDefinitionId = c(10, 10, 10, 10),
+                                subjectId = c(1, 1, 2, 2),
+                                cohortStartDate = c(as.Date("1900-01-01"), as.Date("1900-04-01"), as.Date("1901-01-01"), as.Date("1901-05-01")),
+                                cohortEndDate = c(as.Date("1900-02-28"), as.Date("1900-05-31"), as.Date("1901-02-15"), as.Date("1901-06-30")))
+  
+  cohort <- dplyr::bind_rows(targetCohort, comparatorCohort)
+  
+  DatabaseConnector::insertTable(connection = connection, 
+                                 tableName = "#cohort_relationship",
```
azimov, line 32: I suggest using a permanent table. Temp tables won't always work with this method; the connection that uploads the data is not necessarily the one that is used later (e.g. if bulk upload is used its a different process).

R/CohortRelationship.R
```diff
@@ -0,0 +1,538 @@
...
+        andromedaTableName = 'temp'
+      )
+      
+      if (!"cohortRelationships" %in% names(resultsInAndromeda)) {
+        resultsInAndromeda$cohortRelationships <- resultsInAndromeda$temp
+      } else {
+        Andromeda::appendToTable(resultsInAndromeda$cohortRelationships,
+                                 resultsInAndromeda$temp)
+      }
+      readr::write_excel_csv(
```
azimov, line 344: this is what is causing github actions to fail. Test code on local first. This is easy with devtools::test() which only runs the sqlite tests for this package so is very quick. You can also run just the tests for this PR with 

```
devtools::test(filter="moduleCohortRelationship")
```

azimov (dismissed):

R/CohortRelationship.R
gowthamrao: thats a good tip!

tests/testthat/test-moduleCohortRelationship.R
gowthamrao: Yes - but i just want the test to work from end to end first

tests/testthat/test-moduleCohortRelationship.R
gowthamrao: Is the concern just for testing or is this a concern that needs to be addressed everywhere?

azimov (approved):

PR #759, opened 2022-02-10
Merge hotfix 2.2.3 develop

azimov:


inst/shiny/DiagnosticsExplorer/global.R
```diff
@@ -1,158 +1,315 @@
...
-  prettyAnalysisIds <- specifications$analysisId

-} else {

-  prettyAnalysisIds <- c(0)

-}

-
+library(magrittr)
```
gowthamrao, line 1: what happened here

gowthamrao (changes requested):
Something happened to global.R - do you think u can fix that?

inst/shiny/DiagnosticsExplorer/global.R
azimov: It looks like In a previous commit windows line endings got inserted (the ^M character)  I've seen this happen before with different IDEs, but this got interpreted as an extra line ending and hence why it was messed up. It should be fixed now.

gowthamrao (approved):

PR #762, opened 2022-02-17
Patch to support VA CDM (extra columns that are not in the standard CDM)

azimov:
@msuchard it would be good to be able to add some test cases that mimic the VA's db that has extra columns. I think @gowthamrao has already included [this](https://github.com/OHDSI/CohortDiagnostics/blob/153e48b6b4c1aac056204ce62763615e65e9f542/R/Private.R#L121), but we should definitely confirm this behaviour is supported going forwards.


tests/testthat/test-ResultsDataModel.R
```diff
@@ -49,7 +49,17 @@ test_that("Create schema", {
 
 test_that("Results upload", {
   skip_if(skipResultsDm | skipCdmTests, 'results data model test server not set')
-
+  if (dbms == "sqlite") {
+    # Checks to see if adding extra OMOP vocab, unexpectedly breaks things
+    connection <- DatabaseConnector::connect(connectionDetails)
+    with_dbc_connection(connection, {
+      DatabaseConnector::renderTranslateExecuteSql(connection, "
+        INSERT INTO main.vocabulary
```
azimov, line 57: @msuchard I added this test because, looking at the branch, it seems that the VA's CDM is violating the primary key constraint of the vocabulary table?

I'm hesitant to support this in general - I'm not sure why one CDM would have multiple OMOP standards - but I don't think this will have any negative impacts for CohortDiagnostics

R/Private.R
```diff
@@ -195,13 +195,17 @@ makeDataExportable <- function(x,
...
   
-  if (nrow(x) >  nrow(x %>% dplyr::select(dplyr::all_of(primaryKeyInDataModel)) %>% dplyr::distinct())) {
-    stop(
-      " - duplicates found in primary key for table ",
-      tableName,
-      ". The primary keys are: ",
-      paste0(primaryKeyInDataModel, collapse = ", ")
-    )
+  # check to see if there are primary key collision in tables that have this unique constraint
+  if (length(primaryKeyInDataModel) > 0) {
```
azimov, line 199: @gowthamrao This change just lets us have tables specified in the the DDL without a primary key. I don't see any other way to support the VA's vocabulary table which seems to have multiple CDM version in one somehow.

tests/testthat/test-ResultsDataModel.R
msuchard: I'm not going to support or refute the merit of what the VA VINCI team is doing with their `cdm_source` table.   Their `vocabulary` tables are actually views; this may be part of the issue.

Nonetheless, their `cdm_source` table does contain multiple rows, upon which `CohortDiagnostics` currently fails.  So, if we want `CohortDiagnostics` be less brittle and keep friends happy ....

R/Private.R
gowthamrao: can we just do a max

tests/testthat/test-ResultsDataModel.R
gowthamrao: if its only cdm_source - why dont we just do a max for this one table? 

we dont need the info from this table for "diagnostics"

gowthamrao: that way we dont have to second guess or make any changes to the other areas of the code

R/MetaDataDiagnostics.R
```diff
@@ -164,4 +166,7 @@ getVocabularyVersion <- function(connection, vocabularyDatabaseSchema) {
     dplyr::rename(vocabularyVersion = .data$vocabularyVersion) %>%
     dplyr::pull(.data$vocabularyVersion) %>%
     unique()
+
+  # Edge case where a CDM has more than a single entry
+  paste(vocabularyVersion, collapse = ";")
```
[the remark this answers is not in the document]
azimov: I agree, its a violation of the con straints but this is what the VA are doing so we'll need to support it. This won't have a major impact in any of the cohort diagnostics but it will tell us that users have a funky CDM.

R/MetaDataDiagnostics.R
```diff
@@ -129,6 +129,8 @@ saveDatabaseMetaData <- function(databaseId,
                                  vocabularyVersion) {
   ParallelLogger::logInfo("Saving database metadata")
   startMetaData <- Sys.time()
+  vocabularyVersion <- paste(vocabularyVersion, collapse = ";")
```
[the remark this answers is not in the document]
azimov: This is true, but they aren't following the standard or constraints and I can't think of a better solution

R/Private.R
azimov: Max won't tell us what versions they are potentially using in a cdm

R/MetaDataDiagnostics.R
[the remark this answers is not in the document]
gowthamrao: Can we please confirm that? I am fine with non conformance to cdm_source table - as it doesnt have conformance standards that are clear. But if we make exception to vocabulary table - then we are violating some of the basic tenants of OMOP

R/MetaDataDiagnostics.R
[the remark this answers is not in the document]
gowthamrao: The output of "SELECT * from vocabulary" should not violate the conformance standards expected by OMOP. Can we confirm if this is TRUE i.e. new fields are being added to the vocabulary table that is a primary key violation?

again - i am fine with cdm_source table, but this vocabulary table is making me uneasy.

inst/settings/resultsDataModelSpecification.csv
```diff
@@ -228,8 +228,8 @@ visit_context,visit_concept_id,int,Yes,Yes,No,Yes,No,No,No
 visit_context,visit_context,varchar,Yes,Yes,No,Yes,No,No,No
 visit_context,subjects,float,Yes,No,No,Yes,Yes,No,No
 visit_context,database_id,varchar,Yes,Yes,No,Yes,No,No,No
-vocabulary,vocabulary_id,varchar(50),Yes,Yes,No,Yes,No,Yes,No
+vocabulary,vocabulary_id,varchar(50),Yes,No,No,Yes,No,Yes,No
```
gowthamrao, line 231: This is a standard vocabulary table - shipped by OMOP. This change is making me uncomfortable.

azimov: I agree, but I can't think of any other way to support the VA

R/MetaDataDiagnostics.R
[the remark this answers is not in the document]
gowthamrao: After discussion with @azimov i think it is fine

gowthamrao (approved):

PR #766, opened 2022-02-19
Concept diagnostics

gowthamrao:
Optimize concept set diagnostics

This PR replaces the older concept set diagnostics (that had a few bugs and surprises) to an optimized version

- in this PR, the movement of files from db to local csv is minimized
- bulk orphan logic that performs it for all concept set
- addition of exclusion module


R/CohortComparisonDiagnostics.R
```diff
@@ -67,18 +67,15 @@ computeCohortOverlap <- function(connectionDetails = NULL,
     return(tidyr::tibble())
   }
   
-  sql <- SqlRender::loadRenderTranslateSql(
-    "CohortOverlap.sql",
-    packageName = utils::packageName(),
-    dbms = connection@dbms,
+  sql <- SqlRender::readSql(system.file("sql/sql_server/CohortOverlap.sql",
```
azimov, line 70: instead of slashes use separate strings e.g.

```
system.file("sql", "sql_server", CohortOverlap.sql", package = utils::packageName())
```

R/CohortLevelDiagnostics.R
```diff
@@ -42,20 +42,18 @@ getCohortCounts <- function(connectionDetails = NULL,
...
-  sql <-
-    SqlRender::loadRenderTranslateSql(
-      sqlFilename = "CohortCounts.sql",
-      packageName = utils::packageName(),
-      dbms = connection@dbms,
-      cohort_database_schema = cohortDatabaseSchema,
-      cohort_table = cohortTable,
-      cohort_ids = cohortIds
-    )
+  sql <- SqlRender::readSql(system.file("sql/sql_server/CohortCounts.sql",
```
azimov, line 45: again, remove use of slashes in file paths

R/IncidenceRates.R
```diff
@@ -56,15 +56,13 @@ getIncidenceRate <- function(connectionDetails = NULL,
...
   
   ParallelLogger::logInfo("Calculating incidence rate per year by age and gender")
-  sql <-
-    SqlRender::loadRenderTranslateSql(
-      sqlFilename = "GetCalendarYearRange.sql",
-      packageName = utils::packageName(),
-      dbms = connection@dbms,
-      cdm_database_schema = cdmDatabaseSchema
-    )
+  sql <- SqlRender::readSql(system.file("sql/sql_server/GetCalendarYearRange.sql",
```
azimov, line 59: Code changes in this file are unrelated to the PR. Doing this increases the chances of a bug entering in the codebase

R/ResultsDataModel.R
```diff
@@ -255,12 +255,16 @@ createResultsDataModel <- function(connection = NULL,
...
     stop("Invalid schema for sqlite, use schema = 'main'")
   }
-
-  sql <- SqlRender::loadRenderTranslateSql(sqlFilename = "CreateResultsDataModel.sql",
-                                           packageName = utils::packageName(),
-                                           dbms = connection@dbms,
-                                           results_schema = schema)
-  DatabaseConnector::executeSql(connection, sql)
+  
+  sql <- SqlRender::readSql(system.file("sql/sql_server/CreateResultsDataModel.sql",
```
azimov, line 259: Again, this refactoring is unnecessary and will probably not be tested during development. Its not a major change but why add it here?

R/ResultsDataModel.R
```diff
@@ -531,6 +535,16 @@ deleteAllRecordsForDatabaseId <- function(connection,
                                           schema,
                                           tableName,
                                           databaseId) {
+  tableExists <-
```
azimov, line 538: There are no guarantees this will work and it doesn't seem related to the PR

inst/sql/sql_server/CreateResultsDataModel.sql
```diff
@@ -31,95 +37,42 @@
...
-
-IF OBJECT_ID('@results_schema.@time_distribution', 'U') IS NOT NULL
-    DROP TABLE @results_schema.@time_distribution;
-
-IF OBJECT_ID('@results_schema.@visit_context', 'U') IS NOT NULL
-    DROP TABLE @results_schema.@visit_context;
-
-IF OBJECT_ID('@results_schema.@vocabulary', 'U') IS NOT NULL
-    DROP TABLE @results_schema.@vocabulary;
+DROP TABLE IF EXISTS @results_schema.@cohort;
```
azimov, line 40: This is postgres specific syntax and will break most databasae platforms. Its also not related to the scope of the PR.

inst/sql/sql_server/OptimizeConceptSet.sql
```diff
@@ -0,0 +1,195 @@
+DROP TABLE IF EXISTS #o_not_excluded;
```
azimov, line 1: Again, drop table if exists is postgres specific syntax. This will break most for most platforms

R/IncidenceRates.R
gowthamrao: It can - but we can test it. Reason i am making the change here is to ensure we have the same pattern everywhere. Consistent pattern will help us debug. This pattern is being used everywhere

R/ResultsDataModel.R
gowthamrao: So the reason is 

`SqlRender::loadRenderTranslateSql` has the argument dbms. The default value is "sql_server". But notice that in prior code there is dbms = connection@dbms . So when connection was to sqlite it was looking for code in the path sqlite instead of sql_server. 

I found it confusing as it is an implicit and not explicit. This was implicit, because we always use the sql_server as the core folder for SQL. 

This code change is to make sure we are explicit and consistent in our pattern everywhere in the package (i.e. not just in the code for this PR).

R/ResultsDataModel.R
gowthamrao: I can take that off - thank you.

ofcourse if there is no table - the sql will fail. we just have to make sure that deleteAllRecordsForDatabaseId is not used without first knowing there is such a table.

inst/sql/sql_server/CreateResultsDataModel.sql
gowthamrao: Martijn dropped this new knowledge for me - its supported sqlRender 1.9+

image

R/ResultsDataModel.R
azimov: > SqlRender::loadRenderTranslateSql has the argument dbms. The default value is "sql_server". But notice that in prior code there is dbms = connection@dbms . So when connection was to sqlite it was looking for code in the path sqlite instead of sql_server.

This is confusing to me; the behaviour of `loadRenderTranslateSql` should look for the `sqlite` folder first then the `sql_server` folder. The reason for this is that there may be code that is specific to `sqlite` for which the translation doesn't work. In which case, you would add folders and sql files specific for that db platform. See the[ common data model ](https://github.com/OHDSI/CommonDataModel/tree/master/inst/ddl/5.4)for an example of this.

R/CohortComparisonDiagnostics.R
azimov: Use of sql_server folder directly is not advised, the way `loadRenderTranslateSql` worked before is correct.

R/ResultsDataModel.R
gowthamrao: But i dont get that same error in develop branch - only in the code for this PR. i dont understand why

This works fine
https://github.com/OHDSI/CohortDiagnostics/blob/develop/extras/tests/TestCodeEunomiaPackageMode.R

azimov: The problem you're having is an interaction between devtools and SqlRender's file loader.

This is caused by the way that devtools test works; the normal mode of testing calls an actual package installed in the system but devtools uses a `load_all()` function. This can be solved by using rstudio for tests or creating a shortcut to the `<PackageRoot>/inst/sql` folder: `<PackageRoot>/sql`. This is resolved in linux/mac systems in the `test/testthat/helpers.R` but the solution used there (creating an sql symbolic link/shortcut when the tests run) does not work on windows because of permissions issues on many user's platforms.

R/ConceptSetDiagnostics.R
```diff
@@ -0,0 +1,2295 @@
...
+                    .data$cohortId,
+                    .data$conceptSetId) %>%
+    dplyr::arrange(.data$uniqueConceptSetId,
+                   .data$cohortId,
+                   .data$conceptSetId)
+  return(conceptSets)
+}
+
+
+mergeTempTables <-
```
azimov, line 721: Looks like the tests are failing with this function

R/ConceptSetDiagnostics.R
```diff
@@ -0,0 +1,2295 @@
...
+  function(connection,
+           tableName,
+           tempTables,
+           tempEmulationSchema) {
+    valueString <-
+      paste(tempTables, collapse = "\n\n  UNION ALL\n\n  SELECT *\n  FROM ")
+    sql <-
+      sprintf("SELECT *\nINTO %s\nFROM (\n  SELECT *\n  FROM %s\n) tmp;",
```
azimov, line 729: what's the reason for using sprintf instead of `renderTranslateExecute`?

R/ConceptSetDiagnostics.R
```diff
@@ -0,0 +1,2295 @@
...
+                           tempEmulationSchema = tempEmulationSchema)
+    DatabaseConnector::executeSql(connection,
+                                  sql,
+                                  progressBar = FALSE,
+                                  reportOverallTime = FALSE)
+    
+    # Drop temp tables:
+    for (tempTable in tempTables) {
+      sql <-
+        sprintf("TRUNCATE TABLE %s;\nDROP TABLE %s;", tempTable, tempTable)
```
azimov, line 744: use `renderTranslateExecuteSql`

R/ConceptSetDiagnostics.R
gowthamrao: @azimov  the mergeTempTables function was never changed 

see

https://github.com/OHDSI/CohortDiagnostics/blob/main/R/ConceptSets.R#L171-L204

R/ConceptSetDiagnostics.R
gowthamrao: Just how it was written the first time this function was written two years ago. it is the version in main branch.

R/ConceptSetDiagnostics.R
gowthamrao: Same comment - any change would be unrelated refactor

PR #774, opened 2022-03-15
Update shiny with reactTable, code simplification and annotation

gowthamrao:
Removes a lot of code lines from shiny

- uses reactable instead of DT. This gives a lot of flexibility and overcomes many draw backs of DT
- add an annotation function, To test that the results data model should have the annotation tables. annotation tables are part of the results data model. If they dont exist, it may be crated using code in extras


inst/shiny/DiagnosticsExplorer/server.R
```diff
@@ -4119,69 +2722,226 @@ shiny::shinyServer(function(input, output, session) {
...
+    
+    data <- getDisplayTableSimple(data = data,
+                                  keyColumns = keyColumns,
+                                  dataColumns = dataColumns)
+    return(data)
+  })
+  
+  #Sign in/Authentication ---------------------------------------------
+  activeLoggedInUser <- reactiveVal(NULL)
+  observeEvent(eventExpr = input$signInModelPopUp,
```
azimov, line 2771: I have a few problems with the implementation of this sign in module:

* It seems to crash the shiny app when bad details are entered

* There must always be a webapi for the shiny apps - this is problematic because that couples this entire shiny app to having some webapi instance 

* This is difficult to check and relates all cohorts to a web api instance - what if the cohorts here don't map to the web api instance anymore? 

* This security model isn't ideal in a production setting - if there is a hack there is an obvious DDOS for our production server - keep sending comments until the postgres db crashes. We have no way of preventing this on our live environments

I don't want this to sound like I don't want annotations I just want to get the security model right. I think behind an organisational firewall that the current implementation is ok (i'd actually argue for having no log in and just allowing it to be edited anonymous wiki style, perhaps the user just having to add their name) but if we allow this on data.ohdsi.org there could be significant problems that need addressing. Lee is busy enough as it is!

inst/shiny/DiagnosticsExplorer/ui.R
```diff
@@ -49,9 +51,15 @@ choicesFordatabaseOrVocabularySchema <- list(
   'Reference Vocabulary' = vocabularyDatabaseSchemas
 )
 
+headerContent <- tags$li(
+  shiny::actionButton(inputId = "signInModelPopUp",
```
azimov, line 55: This should be hidden if the webapi instance doesn't exist

inst/shiny/DiagnosticsExplorer/server.R
```diff
@@ -4119,69 +2722,226 @@ shiny::shinyServer(function(input, output, session) {
...
+                         shiny::modalButton("Cancel"),
+                         shiny::actionButton("signIn", "Sign in")
+                       ),
+                       tags$div(
+                         shiny::textInput(
+                           inputId = "userName",
+                           label = "User name",
+                           width = NULL
+                         ),
+                         shiny::passwordInput(
```
azimov, line 2789: This is bad - we're taking a users password in the shiny app (in plain text) then passing it to the webapi instance.

azimov (changes requested):

azimov (approved):

PR #800, opened 2022-04-12
Shiny app changes

azimov:
Added shinycssloaders for tables and figures.

Changed name of connection details in shiny app because it was overriding user's settings during analysis. 

Removed hard coding of temporal choices in temporal characterization tab - was hiding those selected by the user

### DIFF
--- a/inst/shiny/DiagnosticsExplorer/global.R
+++ b/inst/shiny/DiagnosticsExplorer/global.R
@@ -50,11 +50,11 @@ if (exists("shinySettings")) {
   resultsDatabaseSchema <- "main"
   vocabularyDatabaseSchemas <- "main"
   dbms <- "sqlite"
-  connectionDetails <- DatabaseConnector::createConnectionDetails(dbms = "sqlite", server = sqliteDbPath)
+  shinyConnectionDetails <- DatabaseConnector::createConnectionDetails(dbms = "sqlite", server = sqliteDbPath)
 } else {
   writeLines("Connecting to remote database")
   dbms <- Sys.getenv("shinydbDatabase", unset = "postgresql")
-  connectionDetails <- DatabaseConnector::createConnectionDetails(
+  shinyConnectionDetails <- DatabaseConnector::createConnectionDetails(
     dbms = dbms,
     server = Sys.getenv("shinydbServer"),
     port = Sys.getenv("shinydbPort", unset = 5432),
--- a/inst/shiny/DiagnosticsExplorer/global.R
+++ b/inst/shiny/DiagnosticsExplorer/global.R
@@ -39,8 +39,8 @@ if (enableAuthorization) {
 
 if (exists("shinySettings")) {
   writeLines("Using settings provided by user")
-  connectionDetails <- shinySettings$connectionDetails
-  dbms <- connectionDetails$dbms
+  shinyConnectionDetails <- shinySettings$connectionDetails
+  dbms <- shinyConnectionDetails$dbms
   resultsDatabaseSchema <- shinySettings$resultsDatabaseSchema
   vocabularyDatabaseSchemas <- shinySettings$vocabularyDatabaseSchemas
   enableAnnotation <- shinySettings$enableAnnotation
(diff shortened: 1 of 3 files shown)


inst/shiny/DiagnosticsExplorer/global.R
```diff
@@ -187,12 +187,6 @@ if (exists("database")) {
 
 if (exists("temporalTimeRef")) {
   temporalCharacterizationCovariateChoices <- get("temporalTimeRef") %>%
-    dplyr::filter((.data$startDay == -365 & .data$endDay == -31) |
```
azimov, line 190: I'm not sure if this hard coding was intentional or not but it hides user settings that are likely important

gowthamrao: It was intentional. This makes the output match prior versions

gowthamrao: I disagree with opening up the diagnostics explorer to more temporal choices. We deliberately chose the time windows here. If we take this out, there is no difference between temporal-diagnostics and non temporal diagnostics as we have framed it.

inst/shiny/DiagnosticsExplorer/ui.R
```diff
@@ -1238,18 +1238,20 @@ bodyTabItems <- shinydashboard::tabItems(
                                            )
                                          )
                               ),
-                              reactable::reactableOutput("compareCohortCharacterizationTable")),
+                              shinycssloaders::withSpinner(reactable::reactableOutput("compareCohortCharacterizationTable"))),
```
gowthamrao, line 1241: i am fine with adding css loaded spinner , but i want to see how to interacts with progress messages

progress messages - as implemented tell us more than just loading a page. it tell if there is a database query going on vs is R doing a post processing of the data etc.

gowthamrao (changes requested):
We should have temporal restriction to temporalCovariateChoices for the temporalCharacterization and compare temporal characterization diagnostics menu 

If we want to support a new menu page with other time windows that fine by me.. just like we have characterization diagnostics with different time window

gowthamrao:
I think the confusion is how we allow the flexible use of feature extraction covariate settings as input, but limit the output to pre-specified analysis and time windows

we have one output from feature extraction and we are splitting it into FIVE diagnostics

1) time distribution
2) cohort characterization (which has a specific set of pre-defined time windows)
3) compare cohort characterization (Same time windows as 2)
4) temporal characterization (different time windows compared to 2)
5) compare temporal characterization (Same as 4)

gowthamrao:
We can make a design decision to not expose the temporalCovariateSettings

https://github.com/OHDSI/CohortDiagnostics/blob/3c70224cbf32635c7734373766a42bfa8825ba3f/R/RunDiagnostics.R#L141-L192

Or make it a file that is read in - instead of explicit option

inst/shiny/DiagnosticsExplorer/global.R
```diff
@@ -39,8 +39,8 @@ if (enableAuthorization) {
 
 if (exists("shinySettings")) {
   writeLines("Using settings provided by user")
-  connectionDetails <- shinySettings$connectionDetails
```
gowthamrao, line 42: This change has broken the ability to launch shiny app by using the launchApp button in the R studio UI

gowthamrao: image

azimov: Should be fixed in latest push

inst/shiny/DiagnosticsExplorer/global.R
azimov: but we let the users choose these settings. It would *really really* annoy me if I made principled choices for, say, vaccine adverse events at risk windows (which takes time and discussion) then generated the results (which takes longer because of these settings) then I was unable to view them

gowthamrao: Yup. So these are conflicting ideas. 

The solution i have in my mind should work. Will show you

gowthamrao:
after discussing with @azimov will move the code from here to https://github.com/OHDSI/CohortDiagnostics/tree/shinyFixes and do a PR

PR #849, opened 2022-06-30
Inclusion rule display changes

gowthamrao:
This change shows percent of attrition in table instead of raw numbers (that are harder to interpet). plus the table is sorted by inclusion rule id order


azimov:
included in #860

PR #853, opened 2022-07-09
Database migration utilty

azimov:
This PR is designed to allow upgrading old schemas (version 3.0.0 and above) to new database formats as it means existing datasets can utilize new shiny code without having to support multiple data model versions.

Basically this looks for any files inside the `sql/sql_server/migrations` directory that match the pattern `Migration_<int>-MigrationName.sql` and then executes them.
These scripts are executed in order.
The syntax for altering columns isn't well supported by SqlRender or Sqlite so there will likely need to be custom scripts for different dbmses. This is supported. 

Down migrations are not included, and nothing happens inside a transaction. So it is assumed that users will have backups of data.

This does not currently include code to update schemas created from `resultsDataModelSpecification.csv` files (e.g. those created by ASSURE projects) because I need to figure out a way of tracking which migrations have been completed (e.g. some other way of storing a data model version).

This can likely be accomplished by changing the creation sql script to use the csv data file and then storing the last migration index - all migrations can then be added to the migrations table allowing the migrations to be future proof.

Note: I would prefer to use something like [flyway](https://flywaydb.org/documentation/). However, this does not fit niceley into our stack in an obvious way that I could see.


azimov:
This has now been tested on real data generated with version 3 in an sqlite database.

R/DatabaseMigrations.R
```diff
@@ -0,0 +1,104 @@
...
+  tables <- DatabaseConnector::getTableNames(connection, schema)
+  if (toupper(paste0(tablePrefix, "migration")) %in% tables) {
+    migrationsExecuted <- getExecutedMigrations(connection, schema, tablePrefix)
+    migrationsToExecute <- setdiff(allMigrations, migrationsExecuted$migrationFile)
+  } else {
+    migrationsToExecute <- allMigrations
+  }
+
+  ParallelLogger::logInfo("Executing database migrations on schema - ", schema)
+  migrationsToExecute <- sort(migrationsToExecute)
```
chrisknoll, line 71: Can you please confirm that `sort()` will handle the case where you have `Migration-1`, `Migration-11`, `Migration-2` and you get the result: `Migration-1`, `Migration-2`, `Migration-11` and not `Migration-1`, `Migration-11`, `Migration-2`.   If `sort()` is doing an alphabetical sort, then you'll get an out-of-order migration with 11 coming before 2.

chrisknoll (changes requested):

chrisknoll:
Overall, this is very cool functionality (re: the database migration). I have a few suggestions that you can decide how to address them, if at all:

#### R6 Classes

I think defining an R6 class to encapsulate the migration functionality would be useful here.  It would let you encapsulate your migration settings (such as the format of the regex files, the table name to store the migration metadata, and other runtime-paramaters.   It would look something like this:
```
dbMigration<- DBMigraton::MigrationManager$new(tablePrefix="xxx", metaTable="yyy","migrationFolder="zzz"...)
dbMigration$status() # this can report the pending migration
dbMigration$migrate() # migrate the database
```

My thought is that by encapsulating the migration paramaters into an R6 instance, you don't have to repeat those paramater values for each MigrationManager function.   Also, it would allow you to instantiate multiple managers in one application context in the event that you need to manage a set of tables via different migration folders (but as I think of this, I think a single application should be managed by a single migration folder, but this at least gives you flexibility)

#### Other Functions

I think that a `status()` function would be useful so that you can see what migrations are pending based on what is currently in the db meta table vs. what is in the migration directory.

A `check()` function may be useful to verify that the migration folder only has files that conform to the specified naming convention of the folders. In other work I've done, one of the things that bit me was missing a special character in the filename convention, and being able to do a `check()` that reports anything out of order would have helped.

#### Stamd-alone package

I think the migration function would work well as a stand-alone package.  Application packages can decide if they want to invoke your migrations themselves (as you are doing in CohortDiagnostics::migrateDatamodel()) or the package could reveal the migration folder as a package-method, and an administrator can call database migration functions external to the package.

R/DatabaseMigrations.R
azimov: This is a great spot, thanks for the review!

azimov:
@chrisknoll thanks for the detailed review.

I have decided that a standalone package will be the best solution going forward, however I don't have the bandwidth to implement it right now. I think this should also be up for a bit more discussion with the wider hades devs - I believe that if we go this way all packages that use shiny apps should conform to a data model and therefore the migration strategy. However, this should be discussed first as many packages still just output 

The R6 class should be implemented in the standalone package as I agree that this is a better design. My main concern is getting these changes in to version 3.1.0 - I don't feel that adding these functions add much technical debt as this can easily be refactored when the migration package is created.

azimov:
@chrisknoll The code is ready for you to review in full. Some of the tests appear to fail intermittently - they seems to be taking a long time to do this so I think this is a github actions issue.

inst/sql/sql_server/migrations/Migration_1-v3_1_0_time_id.sql
```diff
@@ -0,0 +1,25 @@
...
+{DEFAULT @temporal_covariate_value_dist = temporal_covariate_value_dist}
+{DEFAULT @incidence_rate = incidence_rate}
+{DEFAULT @table_prefix = ''}
+
+-- Create table indicating version number of ddl
+DROP TABLE IF EXISTS @database_schema.@table_prefix@package_version;
+
+--HINT DISTRIBUTE ON RANDOM
+CREATE TABLE @database_schema.@table_prefix@package_version (
+    version_number VARCHAR PRIMARY KEY
```
chrisknoll, line 20: This becomes varchar(MAX) in some platforms, and I wonder if you should apply a max size (like varchar(50) on this.

inst/sql/sqlite/migrations/Migration_2-v3_1_0_ir_person_years.sql
```diff
@@ -0,0 +1,26 @@
...
+-- ALTER TABLE @results_schema.@table_prefix@incidence_rate ALTER COLUMN person_years FLOAT;
+
+ALTER TABLE @database_schema.@table_prefix@incidence_rate RENAME TO _incidence_rate_old;
+
+CREATE TABLE @database_schema.@table_prefix@incidence_rate (
+    cohort_count BIGINT NOT NULL,
+    person_years NUMERIC,
+    gender VARCHAR,
+    age_group VARCHAR,
+    calendar_year VARCHAR(4),
```
chrisknoll, line 15: Same comment about an unbounded VARCHAR, and should calendar year be varchar?  Not only is an int smaller to store, but it would also prevent the storage of things like 'NA' or 'xxxx'

azimov: I'm not sure about the reason these are VARCHAR - it dates back to a design decision made before I was working on this (he're I literally copy pasted the table definition for sqlite). I assume this is because of how the data was stored. I'd suggest moving away in a future version, for now I have just bounded them to values.

PR #863, opened 2022-07-20
basic index event tests + edited visit context tests 

pranavbhimani:


azimov (approved):
Awesome, our test coverage for the shiny app is going up exponentially!

Issue #838, opened 2022-06-27
Logging improvements


azimov:
Currently, logging of the execution of cohort diagnostics is limited and it is difficult to understand which steps are taking a significant amount of time.

The output for each task should be captured in a way that allows granularity at the following levels:

* Total execution for all cohorts and all tasks, per database
* Time per each task for all cohorts
* Time per each cohort for each individual diagnostic task

Capturing these results in a manner which allows profiling afterwards is also required, therefore a csv file (or files) should be output that allows filtering and exploration. It is not currently a requirement that this file is explorable in the diagnostics explorer.

azimov:
Now Included in develop branch

PR #864, opened 2022-07-21
Improved logging

azimov:
This PR aims to improve logging.

All tasks are recorded in an appended csv file that stores the start time of the task, how long it ran, the name of the task (which relates to the function) and the parent calling task.

Not all code is timed - just things that are likely to result in long executions.
The individual cohort id or concept set id can be stored where this is relevant.

Closes. #838


R/CohortCharacterizationDiagnostics.R
```diff
@@ -39,19 +40,25 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
...
-        cohortDatabaseSchema = cohortDatabaseSchema,
-        cdmVersion = cdmVersion,
-        cohortTable = cohortTable,
-        cohortId = cohortIds[start:end],
-        covariateSettings = covariateSettings,
-        aggregated = TRUE
-      )
-
+    timeExecution(
+      exportFolder,
```
gowthamrao, line 44: since HADES convention is to use named arguement - maybe this one was missed?

https://ohdsi.github.io/Hades/codeStyle.html#Use_named_arguments

R/CohortRelationship.R
```diff
@@ -400,21 +400,28 @@ executeCohortRelationshipDiagnostics <- function(connection,
...
-        comparatorCohortIds = cohortDefinitionSet$cohortId,
-        relationshipDays = dplyr::tibble(
-          startDay = temporalStartDays,
-          endDay = temporalEndDays
-        )
-      )
-
+    timeExecution(
+      exportFolder,
+      "runCohortRelationshipDiagnostics",
```
gowthamrao, line 405: named arguments maybe here too

R/ConceptSets.R
```diff
@@ -387,250 +386,181 @@ runConceptSetDiagnostics <- function(connection,
...
-  #                                                concept_id_table = conceptIdTable,
-  #                                                concept_sets_table = "#inst_concept_sets",
-  #                                                progressBar = FALSE,
-  #                                                reportOverallTime = FALSE)
-  # }
+    conceptSets[!duplicated(conceptSets$uniqueConceptSetId),] %>%
+      dplyr::select(-.data$cohortId, -.data$conceptSetId)
+
+  timeExecution(
+    exportFolder,
```
gowthamrao, line 393: maybe missed here too

R/Private.R
```diff
@@ -308,4 +308,28 @@ getPrefixedTableNames <- function(tablePrefix) {
   }
 
   return(resultList)
+}
+
+#' Internal utility function for logging execution of variables
+timeExecution <- function(exportFolder,
```
gowthamrao, line 314: this is amazing! maybe it should be in all OHDSI R pacakges

gowthamrao (approved):

R/Private.R
gowthamrao: ParallelLogger?

gowthamrao:
@azimov  would it be possible to add a column into the log that specifies the databaseId. 

Reason:
- we can append logs from multiple datasources/data partners and compare
- many times when i run CD in parallell over many data sources - the log does not tell me the databaseId anywhere. I have been using a workaround by encoding the databaseId in the file path.

azimov:
>  many times when i run CD in parallell over many data sources - the log does not tell me the databaseId anywhere. I have been using a workaround by encoding the databaseId in the file path.

The database id is not exposed properly in a lot of functions and it doesn't seem worthwhile to me. The purpose of these logs is to be per-database and the output is in the folder for this database. I do not want to write or support any R or Shiny utilities to work with these logs. I expect those viewing the logs to be able to use excel, all this is for is to give an indication of how long tasks are taking in more complex pipelines than those ran on user machines.

PR #867, opened 2022-07-27
Refactored characterization export to use common function for export …

azimov:
…of csv files

Also added test cases with continuous variables - things were missed in previous tests

Added tests for characterization utility in more isolation

Removed unused export andromeda to csv function

Added function to get default covariate settings

Added option to test to use full covariate settings


gowthamrao (changes requested):
Could we please add chunking to writeToCsv similar to [writeAndromeda legacy function)](https://github.com/OHDSI/CohortDiagnostics/blob/5567b89bd29ae84d9719e0805b1f3ee8b516dd9c/R/Incremental.R#L197-L203)

gowthamrao (dismissed):
Thats awesome work @azimov

gowthamrao (approved):

PR #876, opened 2022-08-03
Fix login

azimov:


gowthamrao:
Testing this on my local - connected to remote postgres. with 

> diagExpEnv$enableAnnotation <- FALSE
> diagExpEnv$enableAuthorization <- FALSE
> diagExpEnv$activeUser <- NULL

When i click on 'Concepts in data source' i get this

image



On R console i see this

```
SELECT *
                            FROM assure_phenotypes_o.annotation
                            WHERE annotation_id IN ();
Warning: Error in ._jobjRef_dollar: no field, method or inner class called 'parent' 
```

looks like app is trying to enable annotation even when its switched off

gowthamrao:
Testing with

#### Not recommended outside of secure firewalls deployments
diagExpEnv$enableAnnotation <- FALSE
diagExpEnv$enableAuthorization <- TRUE
diagExpEnv$activeUser <- NULL


I was able to login using credentials. That means - this issue appears to be fixed.

gowthamrao:
I however see this in R console

> This Font Awesome icon ('user-circle') does not exist:
> * if providing a custom `html_dependency` these `name` checks can 
>   be deactivated with `verify_fa = FALSE`

gowthamrao:
Using 

`diagExpEnv$activeUser <- 'anonymous'`

That worked and user: annonymous was logged in on app start.

(not sure if this is useful though. Especially because the app will force this account to be logged in, and since we dont have a log out functionality - no other user can login)

gowthamrao (changes requested):
It mostly worked

there is one warning message on console

found other related issues - described above

gowthamrao:
The message error: no field parent 
is coming from the function local function renderTranslateQuerySql

PR #894, opened 2022-08-05
Use group by across index event counts

### LINKED ISSUE #871: makeDataExportable function reports violation with index event breakdown
Observed

```
Stack trace:
8: stop(" - duplicates found in primary key for table ", tableName, ". The primary keys are: ", paste0(primaryKeyInDataModel, collapse = ", "))
7: makeDataExportable(x = data, tableName = "index_event_breakdown", minCellCount = minCellCount, databaseId = databaseId)
```

Reason  - unknown

azimov:
Fixes #871

### DIFF
--- a/inst/sql/sql_server/CohortEntryBreakdown.sql
+++ b/inst/sql/sql_server/CohortEntryBreakdown.sql
@@ -10,8 +10,8 @@ IF OBJECT_ID('tempdb..@store_table', 'U') IS NOT NULL
 SELECT domain_table,
   domain_field,
   concept.concept_id,
-	concept_count,
-	subject_count
+	sum(concept_count) as concept_count,
+	sum(subject_count) as subject_count
 {@store} ? {
 INTO @store_table
 } : {
@@ -71,4 +71,6 @@ FROM (
 	
 	) concept_counts
 INNER JOIN @vocabulary_database_schema.concept
-	ON concept_counts.concept_id = concept.concept_id;
+	ON concept_counts.concept_id = concept.concept_id
+
+GROUP BY domain_table, domain_field, concept.concept_id;
--- a/inst/sql/sql_server/CohortEntryBreakdown.sql
+++ b/inst/sql/sql_server/CohortEntryBreakdown.sql
@@ -11,7 +11,7 @@ SELECT domain_table,
   domain_field,
   concept.concept_id,
 	sum(concept_count) as concept_count,
-	sum(subject_count) as subject_count
+	max(subject_count) as subject_count
 {@store} ? {
 INTO @store_table
 } : {
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,8 @@ Bug fixes:
 1. Fixed issue uploading results to postgres db caused by null values in primary key field. 
 Removed constraint to fix issue.
 
+2. Fix for `index_event_breakdown` having duplicate entries where concept is observed in the same domain multiple times
+
 
 CohortDiagnostics 3.0.3
 =======================


gowthamrao:
I think the original SQL when written by @pbr6cornell  or @schuemie  was that data was to be partitioned by domain tables - so the union would not cause the conflict.

gowthamrao:
I suspect the issue is coming when the same concept id is populating both source_concept_id field and the _concept_id (standard) of a domain table. 

That is causing duplication here https://github.com/OHDSI/CohortDiagnostics/blob/5567b89bd29ae84d9719e0805b1f3ee8b516dd9c/inst/sql/sql_server/CohortEntryBreakdown.sql#L39-L69 thus violating primary key checks imposed on version 3. 

Since we did not have these checks in version 2-- we did not see the issue

inst/sql/sql_server/CohortEntryBreakdown.sql
```diff
@@ -10,8 +10,8 @@ IF OBJECT_ID('tempdb..@store_table', 'U') IS NOT NULL
 SELECT domain_table,
   domain_field,
   concept.concept_id,
-	concept_count,
-	subject_count
+	sum(concept_count) as concept_count,
```
gowthamrao, line 13: I think it is more appropriate to do a max for subject count

sum for concept_count
max for subject_count

gowthamrao:
The difference between the two sql pointed above is

```
 INNER JOIN @concept_set_table 
 	ON @domain_concept_id = concept_id 
```

vs
```

 INNER JOIN @concept_set_table 
 	ON @domain_source_concept_id = concept_id 
```


This leads to scenario where the primary keys are the same but the value (count, sum) may be different. So union wont consider the rows as the same.

gowthamrao (changes requested):

inst/sql/sql_server/CohortEntryBreakdown.sql
azimov: After talking to @pbr6cornell I changed this to max. Still not sure what the cause is but the issue appears to be rare and this is probably an adequate fix.

gowthamrao (approved):

abbynewbury:
Hi, I think I am still getting this error. The error report is as follows: Thread: Main
Message:   - duplicates found in primary key for table temporal_covariate_ref. The primary keys are: covariateId
Level:  FATAL
Time:  2023-05-30 17:40:55

Stack trace:
9: h(simpleError(msg, call))
8: .handleSimpleError(function (condition)
{
    if (is(condition, "error")) {
        log("FATA
7: stop(" - duplicates found in primary key for table ", tableName, ". The primary keys are: "
6: makeDataExportable(x = characteristics$covariateRef, tableName = "temporal_covariate_ref",
5: exportCharacterization(characteristics = characteristics, databaseId = databaseId, incremen
4: executeCohortCharacterization(connection = connection, databaseId = databaseId, exportFolde
3: eval(expr)
2: timeExecution(exportFolder, "executeCohortCharacterization", cohortIds, parent = "executeDi
1: executeDiagnostics(cohortDefinitionSet, connectionDetails = connectionDetails, cohortTable

R version:
R version 4.2.3 (2023-03-15)

Platform:
x86_64-pc-linux-gnu

Attached base packages:
- stats
- graphics
- grDevices
- utils
- datasets
- methods
- base

Other attached packages:
- PhenotypeLibrary (3.12.0)
- remotes (2.4.2)
- CohortDiagnostics (3.2.2)
- FeatureExtraction (3.2.0)
- Andromeda (0.6.3)
- dplyr (1.1.2)
- DatabaseConnector (6.2.1)

Let me know if there is a fix that I can implement on my end and thank you!

PR #902, opened 2022-08-12
Fixes some annoying but simple bugs in develop branch

gowthamrao:


inst/shiny/DiagnosticsExplorer/R/StartUpScripts.R
```diff
@@ -408,6 +408,20 @@ initializeEnvironment <- function(shinySettings,
       unique() %>%
       sort()
   }
+  
+  if (!is.null(envir$conceptSets)) {
```
azimov, line 412: What is the reason this is included here? The reason this function exists is because there were hundreds of hacks that changed the global namespace. I want to remove it in a future version as this is bad practice for an app this large.

extras/UploadToDatabase.R
```diff
@@ -46,5 +46,21 @@ for (i in (1:length(listOfZipFilesToUpload))) {
   )
 }
 
+tablesInResultsDataModel <- CohortDiagnostics::getResultsDataModelSpecifications() |>
```
azimov, line 49: Remove native pipe. Its use is inconsistent throughout the package, it requires R 4.1.0 (which is NOT a requirement of the shiny app) and its behaviour is different to they tidyverse's pipe in subtle ways that will be confusing.

azimov (changes requested):
Please split this in to separate PRs so the changes can be tested in isolation.

For my sanity I would like to see the following PRs:

* Changes to the reactive inputs - this is the biggest and the changes to the main DiagnosticsExplorerModule code are very extensive. It is not clear what these changes aim to do and if they even fix the issues. (which aren't stated anywhere)
 
*Changes to the cohort definitions module - there are 600 lines of changes here. There is no stated obvious reason for them and this has enormous potential to cause bugs. It isn't reasonable to expect this to be reviewed as part of a larger PR

* Minor cosmetic changes that do not require much testing

* Inclusion of annotation in modules where it was not previously present

* Lower risk bug fixes that can be quickly tested and approved

extras/UploadToDatabase.R
gowthamrao: Are you looking at the right code? i dont see any native pipes? I removed them last week

PR #908, opened 2022-08-23
Append cd_version to results data model

gowthamrao:
cd_version table was added to develop, but the DDL and results data model wasnt updated

needs a drop for example
https://github.com/OHDSI/CohortDiagnostics/blob/e6b5a7633d1320ca42d80713ce775c187043525f/inst/sql/sql_server/CreateResultsDataModel.sql#L75-L79


azimov:
Rejecting this PR as this change is/will be covered by the database migrations. I am currently working on a generic solution for all hades packages that will implement this.

PR #909, opened 2022-08-27
Chunk cohort time series

gowthamrao:


azimov:
Its not clear what the reason for these changes are or what you mean by "chunked". The tests also fail.

gowthamrao:
> Its not clear what the reason for these changes are or what you mean by "chunked". The tests also fail.

Chunking: break a big something into smaller parts :)

When running this on a large set of cohorts (like phenotype library), its best to break it into chunks of iteration. In each iteration, capture the output from Rs memory and write into file. That way, if for any reason the connection gets closed - the result files are written to drive

PR #910, opened 2022-08-28
Refacatored chunking in feature extraction output

gowthamrao:
When running a large batch - e.g. 100+, previous design would only write to csv and incremental log after all 100+ were complete

This was causing loss of compute during disconnection, failure etc.

This PR makes chunking consistent across CohortDiagnostics

### DIFF
--- a/R/CohortCharacterizationDiagnostics.R
+++ b/R/CohortCharacterizationDiagnostics.R
@@ -22,176 +22,162 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
                                      cohortTable = "cohort",
                                      cohortIds,
                                      cdmVersion = 5,
-                                     covariateSettings,
-                                     batchSize = getOption("CohortDiagnostics-FE-batch-size", default = 5)) {
+                                     covariateSettings) {
   startTime <- Sys.time()
   if (is.null(connection)) {
     connection <- DatabaseConnector::connect(connectionDetails)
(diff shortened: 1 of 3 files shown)


R/CohortCharacterizationDiagnostics.R
```diff
@@ -22,176 +22,162 @@ getCohortCharacteristics <- function(connectionDetails = NULL,
...
+          .data$covariateId,
+          .data$sumValue,
+          .data$mean,
+          .data$sd
+        ) %>%
+        dplyr::mutate(timeId = 0)
     }
-
-    if ("covariatesContinuous" %in% names(featureExtractionOutput) &&
+    if ("covariates" %in% names(results)) {
```
azimov, line 121: Looks like most of these conditional cases aren't tested. Is this because of the features used (e.g. tests before weren't using continuous features) or some other reasons?

Maybe these tests can be improved?

gowthamrao: I am sure it can be improved. Those were not tested before. This is a refactor makes it possible to test them! But i have added the main tests, not adding any more

gowthamrao: I am sure it can be improved. Those were not tested before. This is a refactor makes it possible to test them! But i have added the main tests, dont have bandwidth to add more tests. We can do it at a future test a thon thing

PR #912, opened 2022-09-03
Change charachterization UI

azimov:
Changes included in this branch:

* Loading Temporal Characterization and Compare Chaarcterization with event button rather than change with reactive input which would cause significant delays in the application

* Moved where time windows are selected - the generic approach was a bad UX as in many situations it wasn't clear how this impacted the results

* Fixed inclusion rules table stats

* General tidy up of characterization code

* Better handling of time id setting


gowthamrao:
I suggest we add a slider with min/max (max is not necessary) for pulling down data something like this

https://github.com/OHDSI/CohortDiagnostics/blob/3fd11dc1afd28bf0d2a188d7d9618409e191fb51/inst/shiny/DiagnosticsExplorer/R/ResultRetrieval.R#L212


In 99.9% of use cases - we dont need results with very low proportion and those take a big chunk of data being moved from the db to shiny. We can filter those out and get a better UX

gowthamrao:
BTW - if we decide to add a button to retrieve data, instead of reactivity - i think there are many places we can do that

- incidence rate plot
- cohort overlap plot

gowthamrao:
Examples of tighter spacing between UI elements

Incidence rate plot
image

Cohort Characterization
image

gowthamrao:
> which would cause significant delays in the application

So if we think about this as three distinct steps

1) Data pull from remote db (sqlite or postgres)
2) Post processing retrieved results in R (some additional filtering, transposing, summarizing) that were not in db query
3) Rendering as plot or table

For very large data (several thousands of rows and multiple db)
1) We can reduce data not needed by using filter. I have found this to be very succesful
> In 99.9% of use cases - we dont need results with very low proportion and those take a big chunk of data being moved from the db to shiny. We can filter those out and get a better UX - mostly speed. If a user wants very low count - they can ask for it thru the UI (i.e. maybe a slider that is by default filtering output. ) This design is consistent with current Atlas characterization implementation.

2) There will always be some data post processing - e.g. transposing from long to wide is easier in R than SQL. There is not much opportunity here in my opinion.

3) Rendering very large tables: Reactable does not do server side rendering https://github.com/glin/reactable/issues/22 . So - if the output is large, this will take time and freeze the browser. So - if we want to support very large table, we have to create a way to 'stream' data using some custom rendering on top of regular rendering

gowthamrao:
> Rendering very large tables: 

Continuing on this idea. Instead of doing a custom rendering of reactTable, i found it easier and simpler to just add filters that reduce the payload to reactable. 

Something like this was easy to implement

image



Instead of radio button, a slider would be better i think. Irrespective, the idea here is - we can use native reactTable without the need for custom rendering - by just controlling the payload

azimov:
I agree that the inputs can be closer to save screen space, however, i think the tables are a bit too cluttered in smaller resolutions to I used standard shiny columns. This allows users to see selected options a bit more easily. 

The design for the tabs now looks like this:

Screenshot 2022-09-12 at 7 58 33 AM

(Temporal table)

Screenshot 2022-09-12 at 7 59 45 AM

Comparisons. 

l have also added a button using event reactive for CohortCharacterization tabs butI leave any changes to other tabs for a separate PR/discussion

gowthamrao:
> i think the tables are a bit too cluttered in smaller resolutions

I have seen in lower resolution/full resolution monitors and a phone screen when implemented more tighther like this https://data.ohdsi.org/PhenotypeLibrary

it did not appear cluttered

gowthamrao:
These are cosmetic comments - not critical, but - i dont think it will clutter., The drop downs open up wider as shown here.

image

azimov:
> These are cosmetic comments - not critical
> 
> The drop downs open up wider as shown here. i.e. if u make it tighter, we can save screen white space and it is less likely will be clutter


Ok, I agree we can improve the cosmetics but I'd like to do this in a separate PR - i think we should discuss the changes and then try and standardize these across the different tabs

PR #913, opened 2022-09-08
Cohort Relationship is to computed between first occurrence of target…

gowthamrao:
… and all occurrence of comparator cohort

### DIFF
--- a/R/CohortRelationship.R
+++ b/R/CohortRelationship.R
@@ -127,7 +127,8 @@ runCohortRelationshipDiagnostics <-
             package = utils::packageName()
           )
         )
-      DatabaseConnector::renderTranslateQuerySqlToAndromeda(
+      
+      DatabaseConnector::renderTranslateExecuteSql(
         connection = connection,
         tempEmulationSchema = tempEmulationSchema,
         sql = cohortRelationshipSql,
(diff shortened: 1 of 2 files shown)


azimov:
@gowthamrao this PR seems fine but the description is unclear - is what you're saying that, currently, cohort relationships calculates the overlap based on any time point between the cohorts. However, your desired behaviour is that it's only between the first occurrence?

I think part of the problem with my understanding here is that what the relationships diagnostic test is trying to capture isn't really specified. It would be really helpful to know what you think the function should be doing so I can validate the logic.

gowthamrao:
No - the original (as in current master) and intended behavior is 
- target cohort should be first occurrence
- comparator cohort all occurrence

https://github.com/OHDSI/CohortDiagnostics/blob/befe0eb0e488fa8a4d308fed861cb03c64b67294/R/CohortRelationship.R#L130


When i refactored cohortRelationship to introduce chunking - i missed this firstOccurrence and introduced a bug

so this is a commit to that branch/pull request

azimov:
> 
> When i refactored cohortRelationship to introduce chunking - i missed this firstOccurrence and introduced a bug
> 
> so this is a commit to that branch/pull request

Ok, i follow this will get merged in

PR #937, opened 2022-09-20
New function to subset results

gowthamrao:
This is a function i have found useful. 

Use case: During phenotype development and evaluation we develop and evaluate many candidate cohort definitions. But after the process is complete - we reject many cohorts and just want to remove them. We do not intend them to be shared for various reasons.

This function, takes a previously computed cohort diagnostics output and subsets it to a set of cohortIds given by the user, creates a new zip file that can then be shared.

If this is a useful function, but requires some changes - please feel free to make changes .


R/ResultsDataModel.R
```diff
@@ -598,3 +598,110 @@ deleteAllRecordsForDatabaseId <- function(connection,
...
+      for (j in (1:length(tablesInResultsDataModel))) {
+        if (paste0(tablesInResultsDataModel[[j]], ".csv") %in% listOfFilesInZippedFolder) {
+          dataFromZip <-
+            readr::read_csv(file = file.path(
+              exportDirectory,
+              paste0(tablesInResultsDataModel[[j]], ".csv")
+            ),
+            col_types = readr::cols())
+          
+          if ("cohort_id" %in% colnames(dataFromZip)) {
```
azimov, line 677: I'm not sure this works for all files - what about cohort relationships?

gowthamrao: Should be possible to support it. But you think this function is useful and want to add to the package?

gowthamrao: cohort Relationship has cohort_id - so it should work like any other table correct?

PR #951, opened 2022-09-26
Interval algebra

gowthamrao:


gowthamrao:
The use case motivating this PR is described in a vignette that is part of this PR.

gowthamrao:
After talking to @azimov we will make this standalone package.

PR #957, opened 2022-10-04
concept_id of int instead of big int in data type

### LINKED ISSUE #717: CohortDiagnostics::uploadResults() - CONCEPT_ID as INT
~~~
Uploading table concept
- Preparing to upload rows 1 through 2533
Warning: warning: Problem with `mutate()` column `concept_id`.
i `concept_id = .Primitive("as.integer")(concept_id)`.
i NAs introduced by coercion to integer range
~~~
For DPs that have self generated CONCEPT_IDs in the 3B range it will throw warnings.  Not sure if it still loads data but will continue to run.

Example:
`3000008158`
`3000008132`
`3000008159`

This happens where other CONCEPT_IDs can be:
- CONCEPT_RELATIONSHIP 
- INCLUDED_SOURCE_CONCEPT
- INDEX_EVENT_BREAKDOWN
- ORPHAN_CONCEPT
- VOCABULARY

Note: CUAMC

azimov:
AN issue was found where concept ids were overlflowing in concept set analysis - this aims to fix that bug

Fixes #717

 also, database_id forced to be read as a character to prevent type issues when merging with old data

### DIFF
--- a/R/Incremental.R
+++ b/R/Incremental.R
@@ -266,6 +266,10 @@ saveIncremental <- function(data, fileName, ...) {
       lazy = FALSE
     )
     if ((nrow(previousData)) > 0) {
+      if("databaseId" %in% colnames(previousData)) {
+        previousData$databaseId <- as.character(previousData$databaseId)
+      }
+
       if (!length(list(...)) == 0) {
         idx <- getKeyIndex(list(...), previousData)
       } else {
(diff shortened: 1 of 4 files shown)


gowthamrao:
Fixes this issue? https://github.com/OHDSI/CohortDiagnostics/issues/717

azimov:
> Fixes this issue? #717

Yes actually! We just found a bug with the upload - concept id was set as int in the specs which caused it to become null when it was uploaded. Data transfers should be fixed with this patch

PR #958, opened 2022-10-04
merge branch 'master' into stanford2

eminty:
stanford2 comments out an 'ExportConceptInformation' call in RunDiagnostics.R.  The former calls getTableNames(), which doesn't seem to work at Stanford, which I think is due to a local security setting (as opposed to a broader BigQuery issue).


azimov:
@eminty - these changes are now merged please let me know if you have an problems

PR #1014, opened 2022-12-01
Removing reliance on ggplot internals

DrTTrousers:
In alignment with the changes to gghx (the origin of the code used to facet nested objects in the diagnostics explorer) and to correct an issue found within our team, I have modified the diagnostics explorer code with the facet_nested.R and facet_grid2.R files from https://github.com/teunbrand/ggh4x/tree/master/R.

This prevents several errors with the newer releases of ggplot, as listed in the commit message, and allows the shiny app to launch with the up to date packages.

Using cohort diagnostics most recent version was not possible for us without this change.


inst/shiny/DiagnosticsExplorer/R/FacetGrid2.R
```diff
@@ -0,0 +1,443 @@
...
+    cols = facets$cols,
+    margins = margins,
+    labeller = labeller,
+    as.table = as.table,
+    switch = switch,
+    drop = drop,
+    axes = axes
+  ))
+  
+  ggproto(
```
azimov, line 153: please use calls to external libraries using the `::` namespace (e.g. `ggplot2::ggproto`). 
This is because we don't want to create problems by calling `library` or `require` which will overwrite anything in the user's R session.

inst/shiny/DiagnosticsExplorer/R/FacetGrid2.R
```diff
@@ -0,0 +1,443 @@
...
+  # Check arguments
+  switch <- switch %||% "none"
+  switch <- arg_match0(switch, c("none", "both", "x", "y"))
+  labeller <- check_labeller(labeller)
+  axes  <- .match_facet_arg(axes,   c("margins", "x", "y", "all"))
+  free  <- .match_facet_arg(scales, c("fixed", "free_x", "free_y", "free"))
+  space <- .match_facet_arg(space,  c("fixed", "free_x", "free_y", "free"))
+  rmlab <- .match_facet_arg(rmlab,  c("none", "x", "y", "all"))
+  indy  <- .match_facet_arg(indy,   c("none", "x", "y", "all"))
+  strip <- assert_strip(strip)
```
azimov, line 133: this assertion doesn't exist in the namespace of `ggplot`. is it from `ggh4x`?

azimov (changes requested):
Thanks for the PR @DrTTrousers this looks like it is a useful change.
However, I don't have a major problem adding `ggh4x` to the required packages for the shiny app (installed in the code in `R/Shiny.R` with the [`launchDiagnosticsExplorer`](https://github.com/OHDSI/CohortDiagnostics/blob/1d72a4dd023be145ac775992bc38893535b5351f/R/Shiny.R#L113) function. This will likely be better going forwards as we intend to move this app to `OhdsiShinyModules` package which will have more dependencies anyway.

azimov:
@DrTTrousers the issue your PR fixes has now been resolved -- ggh4x is now a requirement of the package. This will be released in version 3.1.1 later today

PR #1037, opened 2023-02-24
Increase concept.concept_code size in rdms - fixes #1036

### LINKED ISSUE #1036: Increase `concept_code` size in resultsDataModelSpecification.csv
In some cases, the `concept.concept_code` table in the CDM will exceed the 50 character limit. I'd propose to increase the size in the results data model to handle this edge case.

anthonysena:
Updates `inst/settings/resultsDataModelSpecification.csv` to increase the size of `concept.concept_code` per #1036.


azimov (changes requested):
Sorry for the delay on this. This will require a migration for the results schema by adding a migration into the `inst/sql_server/migrations` folder.

See here for an example:

`https://ohdsi.github.io/ResultModelManager/articles/CreatingMigrations.html`

azimov:
should now be resolved in the develop branch so closing this. The column was already varchar max so no migration was needed. The field type is checked but no the size (in any R code) so I don't know where this would have been creating issues, I'd be interested to know more

PR #1092, opened 2023-11-06
Update CreateResultsDataModel.sql

DrTTrousers:
Correcting a bug with CreateResultsDataModel sql where the source concept ID is constrained to Not Null despite being an optional field


azimov (changes requested):
Changes to the results data model require writing migrations to prevent issues when we have to manage historical result sets. 

For examples, see `inst/sql/sql_server/migrations/Migration_2-v3_1_0_ir_person_years.sql`

azimov:
After evaluation this result will not fix the issue - its perfectly valid for the source_concept_id to be null in the cdm schema, but its also valid (and very common) for a standard concept to map to multiple source concepts. Therefore, the primary key will throw an error when the result is imported

PR #1109, opened 2024-03-18
Make data exportable handles empty data frame

azimov:


chrisknoll:
Ok, in my use-case where the cohorts were empty, this runs without error now.  Thank you.

Issue #1067, opened 2023-08-08
export createConceptCountsTable() 


cbarbozaerasmus:
At Darwin EU we are trying to run CohortDiagnostics in a more efficient way when running it for multiple studies. We believe it would be a good enhancement to export the function `createConceptCountsTable()` to generate the `concept_counts` before executing the diagnostics. 

By doing this, we can perform this calculation just once for a specific `vocabulary_version`, instead of repeating this process for each study.

In our fork [darwin-eu-dev/CohortDiagnostics](https://github.com/darwin-eu-dev/CohortDiagnostics), the user can `createConceptCountsTable()`. This table is saved in the `cohortDatabaseSchema`.

```r
CohortDiagnostics::createConceptCountsTable(connectionDetails = connectionDetails,
                                            cdmDatabaseSchema = cdmDatabaseSchema,
                                            conceptCountsDatabaseSchema = cohortDatabaseSchema,
                                            conceptCountsTable = "concept_counts",
                                            removeCurrentTable = TRUE)
```
Then we use the parameter `useExternalConceptCountsTable` in `executeDiagnostics()`. If TRUE, `executeDiagnostics()` uses the `concept_counts` created previously in the `cohortDatabaseSchema`. The user should specify the name of the external concept counts table, generally `concept_counts`

```r
CohortDiagnostics::executeDiagnostics(cohortDefinitionSet =  cohortDefinitionSet,
                                      connectionDetails = connectionDetails,
                                      cohortTable = cohortTable,
                                      cohortDatabaseSchema = cohortDatabaseSchema,
                                      cdmDatabaseSchema = cdmDatabaseSchema,
                                      conceptCountsTable = "concept_counts",
                                      exportFolder = exportFolder,
                                      databaseId = "Eunomia",
                                      minCellCount = 5,
                                      useExternalConceptCountsTable = FALSE)
```
We also modified the `CreateConceptCountTable.sql` file, to add a new column with the vocabulary_version. 

https://github.com/darwin-eu-dev/CohortDiagnostics/blob/ca6d9074bb097b9ce60b7bc6bf72e68a84f650fe/inst/sql/sql_server/CreateConceptCountTable.sql#L102C1-L106C2

```sql
{@table_is_temp} ? {} : { 
ALTER TABLE @work_database_schema.@concept_counts_table
ADD vocabulary_version VARCHAR(20) NULL;
[...]

PR #1117, opened 2024-06-07
Use external concept counts

cebarboza:
Hi, 

I would like to submit this contribution related to [#1067](https://github.com/OHDSI/CohortDiagnostics/issues/1067). 

- Export of `externalConceptCounts()` to create that table in the write schema. 
- `useExternalConceptCountsTable` parameter in runDiagnostics. 
- Changes to R/ConceptSets.R to `useExternalConceptCountsTable` when TRUE or create a new table as default when FALSE.
- Changes inst/sql/sql_server/CreateConceptCountTable.sql to remove the ExternalConceptCountsTable if needed.

Please let me know what you think to see if we can work on this changes!


R/ConceptCountsTable.R
```diff
@@ -0,0 +1,82 @@
...
+#' @param cdmDatabaseSchema CDM schema
+#' 
+#' @return the concepts count table name
+#' @export
+getConceptCountsTableName <- function(connection, cdmDatabaseSchema) {
+  result <- "concept_counts"
+  sql <- paste("SELECT vocabulary_version as version",
+               "FROM @cdmDatabaseSchema.VOCABULARY",
+               "WHERE vocabulary_id = 'None'")
+  dbVersion <- renderTranslateQuerySql(connection = connection,
```
azimov, line 74: Please use DatabaseConnector:: in case a user masks the call in their local environment

R/ConceptCountsTable.R
```diff
@@ -0,0 +1,82 @@
...
+                                       sql = sql,
+                                       cdmDatabaseSchema = cdmDatabaseSchema) %>%
```
azimov, line 76: I'd prefer use of the native pipe `|>` over `%>%` in new code.

R/ConceptCountsTable.R
```diff
@@ -0,0 +1,82 @@
...
+#' @param connection database connection
+#' @param cdmDatabaseSchema CDM schema
+#' @param tempEmulationSchema schema to emulate temp table
+#' @param conceptCountsDatabaseSchema schema name for the concept counts table
+#' @param conceptCountsTable table name
+#' @param conceptCountsTableIsTemp boolean to indicate if it should be a temporary table
+#' @param removeCurrentTable if the current table should be removed
+#'
+#' @export
+createConceptCountsTable <- function(connectionDetails = NULL,
```
azimov, line 31: given that this function is now exported adding some unit tests for basic functionality is probably required

R/ConceptCountsTable.R
```diff
@@ -0,0 +1,82 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#' createConceptCountsTable
+#' 
+#' @description Create a table containing the concept counts.
+#' 
+#' @param connectionDetails database connection details
+#' @param connection database connection
+#' @param cdmDatabaseSchema CDM schema
```
azimov, line 23: for most parameters here it would be better and more consistent to use `#'@inheritParams executeDiagnostics`

R/RunDiagnostics.R
```diff
@@ -710,6 +715,37 @@ executeDiagnostics <- function(cohortDefinitionSet,
       }
     )
   }
+  
+  # Defines variables and checks version of external concept counts table -----
+  if (useExternalConceptCountsTable == FALSE) {
```
azimov, line 720: `!useExternalConceptCountsTable`

R/RunDiagnostics.R
```diff
@@ -710,6 +715,37 @@ executeDiagnostics <- function(cohortDefinitionSet,
...
+    conceptCountsTableIsTemp <- TRUE
+    if (conceptCountsTable != "#concept_counts") {
+      conceptCountsTable <- "#concept_counts"
+    }
+  } else {
+    if (conceptCountsTable == "#concept_counts") {
+      stop("Temporary conceptCountsTable name. Please provide a valid external ConceptCountsTable name")
+    }
+    conceptCountsTableIsTemp <- FALSE
```
azimov, line 729: This flag variable is not needed as it is logically equivalent to `useExternalConceptCountsTable`

R/ConceptCountsTable.R
```diff
@@ -0,0 +1,82 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#' createConceptCountsTable
+#' 
+#' @description Create a table containing the concept counts.
```
azimov, line 19: As this function can save users time, this could be elaborated on somewhat (e.g. some example usage)

azimov (changes requested):
Overall I think this could speed up the phenotyping process quite significantly - using Achilles counts as well somewhat conflicts with this branch. I think that is probably a better approach in general as users should be running this anyway as part of an ETL process.

However, I don't want to hold things up too much. This code should make it into the next release with a few minor changes suggested here.

R/ConceptCountsTable.R
cebarboza: Done

R/ConceptCountsTable.R
cebarboza: Done

R/RunDiagnostics.R
cebarboza: Done

R/ConceptCountsTable.R
cebarboza: I expanded a bit the description.

R/RunDiagnostics.R
```diff
@@ -110,6 +110,8 @@ getDefaultCovariateSettings <- function() {
 #'                                    diagnostics to.
 #' @param cohortDefinitionSet         Data.frame of cohorts must include columns cohortId, cohortName, json, sql
 #' @param cohortTableNames            Cohort Table names used by CohortGenerator package
+#' @param conceptCountsTable          Concepts count table name. The default is "#concept_counts" to create a temporal concept counts table.
```
azimov, line 113: THese changes look good but you need to run `devtools::document()` (or the build in rstudio) to add the changes to  include these parameters.  Doing this should get the build to finish and I can then merge in your changes.

azimov (changes requested):
Two minor changes:

* The tests should be improved to run on all platforms, not just sqlite
* The build fails because the new arguments are not documented yet with devtools::document

I can likely fix the latter one once merged in to develop. However, I would like tests to at least pass before doing this

PRijnbeek:
What is the current status of this? is it added to an upcoming release? and if so when is this release planned (there are no due dates on the milestones in this package)?

ablack3:
> What is the current status of this? is it added to an upcoming release? and if so when is this release planned (there are no due dates on the milestones in this package)?

Current status is that it will be included in the upcoming release. There are a few things left to do to get R check working.

It is part of the 3.4 release - https://github.com/OHDSI/CohortDiagnostics/milestone/3

There is no planned release date yet. @azimov should we set a release date for 3.4?

azimov:
> > What is the current status of this? is it added to an upcoming release? and if so when is this release planned (there are no due dates on the milestones in this package)?
> 
> Current status is that it will be included in the upcoming release. There are a few things left to do to get R check working.
> 
> It is part of the 3.4 release - https://github.com/OHDSI/CohortDiagnostics/milestone/3
> 
> There is no planned release date yet. @azimov should we set a release date for 3.4?

The other changes that are in the 3.4.0 are relatively small and shouldn't take long - we can set a date for a few weeks from now or whenever this PR is finished

azimov:
@ablack3 @cebarboza Any update on this PR? I would like to start testing 3.4.0 next week. If this can make it in it would be great, otherwise we can shelve this for a future date.

ablack3:
Hi @azimov, I'm looking at it now. Your request that we test the createConceptCounts table on all database systems will make the tests take a lot of time I think. Maybe we can use an option to make a small version of the table in tests.

R/ConceptCountsTable.R
ablack3: Shouldn't the concept counts table be created in the cohortDatabaseSchema?

Issue #1118, opened 2024-06-10
$ operator is invalid for atomic vectors


ablack3:
I'm running cohort diagnostics (current OHDSI main branch) and getting the following error.

Thread: Main
Message:  $ operator is invalid for atomic vectors
Level:  FATAL
Time:  2024-06-06 12:36:18

Stack trace:
4: RunDiagnostics.R#807: executeCohortRelationshipDiagnostics(connection = connection, databaseId = databaseId, exportFolder = exportFolder, cohortDatabaseSchema = cohortDatabaseSchema, cdmDatabase
3: Private.R#326: eval(expr)
2: RunDiagnostics.R#801: timeExecution(exportFolder, "executeCohortRelationshipDiagnostics", cohortIds, parent = "executeDiagnostics", expr = {
    executeCohortRelationshipDiagnostics(connection = 
1: CohortDiagnostics::executeDiagnostics(cohortDefinitionSet = cohortDefinitionSet, connectionDetails = connectionDetails, cdmDatabaseSchema = cdmSchema, cohortDatabaseSchema = writeSchema, databas

R version:
R version 4.4.0 (2024-04-24)

Platform:
x86_64-pc-linux-gnu

Attached base packages:
- stats
- graphics
- grDevices
- datasets
- utils
- methods
- base

![image](https://github.com/OHDSI/CohortDiagnostics/assets/10227522/cbac3204-9066-404e-ac19-225dcee3f86e)

Some cohorts have zero counts in my database.

ablack3:
https://github.com/OHDSI/CohortDiagnostics/blob/03c40ab06832a329f2c904fbd686a0596ea9af6f/R/RunDiagnostics.R#L816

`temporalCovariateSettings` is a covariateSettings object but when `executeCohortRelationshipDiagnostics` is called only the first element of the list is passed in. So inside `executeCohortRelationshipDiagnostics` the `temporalCovariateSettings` is a length one logical vector.

Then inside `executeCohortRelationshipDiagnostics` it seems as if `temporalCovariateSettings` is expected to be a a covariate settings object.

https://github.com/OHDSI/CohortDiagnostics/blob/03c40ab06832a329f2c904fbd686a0596ea9af6f/R/CohortRelationship.R#L266


So I think 
https://github.com/OHDSI/CohortDiagnostics/blob/03c40ab06832a329f2c904fbd686a0596ea9af6f/R/RunDiagnostics.R#L816
should not include the subsetting. 

[...]

PR #1119, opened 2024-06-11
remove subsetting of temporalCovariateSettings when running relationship diagnostics

ablack3:
Addresses #1118


azimov:
I think the issue with this PR is that multiple temporal covariate settings can be passed as optional additional analysis (e.g. custom covairates) this would cause relationships/overlap code to fail.

I think the current develop branch should fix your issue.

PR #1120, opened 2024-06-12
add option to use Achilles for concept counts

ablack3:
#1067 

I have not tested this code yet.

TODO
- test the code and add unit tests


R/RunDiagnostics.R
```diff
@@ -232,7 +236,9 @@ executeDiagnostics <- function(cohortDefinitionSet,
                                sampleN = 1000,
                                seed = 64374,
                                seedArgs = NULL,
-                               sampleIdentifierExpression = "cohortId * 1000 + seed") {
+                               sampleIdentifierExpression = "cohortId * 1000 + seed",
+                               useAchilles = FALSE, 
+                               achillesDatabaseSchema) {
```
azimov, line 241: should this be achillesDatabaseSchema = resultsDatabaseSchema?

R/RunDiagnostics.R
```diff
@@ -749,7 +762,9 @@ executeDiagnostics <- function(cohortDefinitionSet,
           useExternalConceptCountsTable = FALSE,
           incremental = incremental,
           conceptIdTable = "#concept_ids",
-          recordKeepingFile = recordKeepingFile
+          recordKeepingFile = recordKeepingFile,
+          useAchilles = useAchilles,
+          resultsDatabaseSchema = resultsDatabaseSchema
```
azimov, line 767: I think this should be `achillesDatabaseSchema`

R/RunDiagnostics.R
```diff
@@ -152,7 +152,11 @@ getDefaultCovariateSettings <- function() {
 #'                                   This expression can use the variables 'cohortId' and 'seed'.
 #'                                   Default is "cohortId * 1000 + seed", which ensures unique identifiers
 #'                                   as long as there are fewer than 1000 cohorts.
-
+#'                                   
+#' @param useAchilles                Logical. Should the pre-computed Achilles analyses be used to get concept counts? TRUE or FALSE (default)
```
azimov, line 156: At least a link to achilles somewhere in the documentation would be great here as not all users will have experience with it.

inst/sql/sql_server/CreateConceptCountTable.sql
```diff
@@ -16,82 +17,152 @@ INTO @concept_counts_table
...
+
+} : {
+
 FROM (
 	SELECT condition_concept_id AS concept_id,
 		COUNT_BIG(*) AS concept_count,
 		COUNT_BIG(DISTINCT person_id) AS concept_subjects
-	FROM @cdm_database_schema.condition_occurrence
+	FROM @cdm_database_schema.condition_occurrence co
+	JOIN @cdm_database_schema.observation_period op
```
azimov, line 53: I don't have a particularly strong opinion on forcing the concepts to occur inside the observation windows (it makes perfect sense for this to be a rule) but this will change the results of the package. However, maybe this should be parameterized to allow users to create consistent reports?

I'd like to hear some thoughts here before making a decisions one way or the other

azimov (changes requested):
I love this approach for using the pre-computed data - only some minor changes.

Also, adding in a unit test (this can be pre-computed counts or a toy example)

inst/sql/sql_server/CreateConceptCountTable.sql
ablack3: My impression is that when people create cohorts with Atlas they are most often using concepts inside the observation period. In the case where a CDM has many concepts outside observation periods it might be confusing to display concepts outside observation period in CohortDiagnostics which are not picked up by Atlas cohort definitions by default. It should be clear that if someone is expecting Atlas to pick up the same number of records they see in CohortDiagnostics they will need include concepts outside observation period. The whole notion of concepts outside observation period is confusing even though I do understand there are legitimate arguments for it. It is hard to explain to OHDSI newcomers: "Observation time is when we see events but there are also events that occur outside observation time that you may want to include". It's strange because if you strictly interpret the definition of observation time there should not be observations outside of it.

R/ConceptSetUtils.R
```diff
@@ -82,12 +82,19 @@ createConceptCountsTable <- function(connectionDetails = NULL,
                                      tempEmulationSchema = NULL,
                                      conceptCountsDatabaseSchema = cdmDatabaseSchema,
                                      conceptCountsTable = "concept_counts",
-                                     conceptCountsTableIsTemp = FALSE) {
+                                     conceptCountsTableIsTemp = FALSE,
+                                     useAchilles = FALSE,
+                                     resultsDatabaseSchema) {
```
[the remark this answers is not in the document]
ablack3: changed.

R/ConceptSetUtils.R
```diff
@@ -82,12 +82,19 @@ createConceptCountsTable <- function(connectionDetails = NULL,
...
   ParallelLogger::logInfo("Creating internal concept counts table")
   if (is.null(connection)) {
     connection <- DatabaseConnector::connect(connectionDetails)
     on.exit(DatabaseConnector::disconnect(connection))
   }
+  
+  if (isTRUE(useAchilles)) {
+    checkmate::assert_character(resultsDatabaseSchema, min.chars = 1, len = 1, any.missing = FALSE)
```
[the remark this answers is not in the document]
ablack3: This is checked at the beginning of executeDiagnostics. I can remove the check here.

R/RunDiagnostics.R
ablack3: resultsDatabaseSchema is not a parameter to executeDiagnostics though. We could set the default to `achillesDatabaseSchema = cdmDatabaseSchema`

R/RunDiagnostics.R
ablack3: changed.

R/RunDiagnostics.R
ablack3: added.

ablack3:
changes made. I moved this PR into the v3.3.0 branch.

ablack3:
@azimov I finally relized that we don't have person counts by source concept id in achilles. Overlooked that before. So until that is available in Achilles I guess we should go with @cebarboza's PR that uses a pre-existing concept counts table. I'll repoen that PR.

PR #1132, opened 2024-09-04
Fix subset conceptset analysis

ganisimov:
This fixes concept set analysis for subset cohorts. Concept sets are defined in parent cohorts, so that first we need to look for parent cohort. That was done, but then parent's cohort sql/json wasn't used - this is fixed.

I'm targeting main branch since the change is trivial and useful for what we're doing now. Please let me know if develop branch is preferred.

### DIFF
--- a/R/ConceptSets.R
+++ b/R/ConceptSets.R
@@ -833,6 +833,18 @@ runConceptSetDiagnostics <- function(connection,
         }
       }
 
+      if (nrow(data) == 0 && ncol(data) == 0) {
+        data <- dplyr::tibble(
+          conceptId = numeric(),
+          conceptCount = numeric(),
+          subjectCount = numeric(),
+          cohortId = numeric(),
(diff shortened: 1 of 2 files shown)


azimov (dismissed):

ganisimov:
Hi @azimov! I was looking into why tests fail. The reason is that subset cohort which was skipped due to the issue is now processed, but has no index event breakdown data. So at the line below the `data` variable holds empty `tibble()` object without any columns:
https://github.com/ganisimov/CohortDiagnostics/blob/a81397212ea797a3c143f8c87cf14283acc07c66/R/ConceptSets.R#L800

This makes subsequent `makeDataExportable()` to call `stop()` due to missing required columns in `data`.

To me it looks like in this case we should just add columns and save empty CSV with proper column names. Please, let me know if you think this is a reasonable solution, I'll add the change to make tests to pass.

PR #1135, opened 2024-09-17
Tidy cohort relationships

azimov:
Adds migrations to remove unused tables and columns

### DIFF
--- a/inst/sql/sql_server/CohortRelationship.sql
+++ b/inst/sql/sql_server/CohortRelationship.sql
@@ -24,66 +24,36 @@ SELECT t.cohort_definition_id cohort_id,
 				THEN c.subject_id
 			ELSE NULL
 			END) sub_cs_before_ts,
-	COUNT_BIG(DISTINCT CASE 
-			WHEN c.cohort_start_date < DATEADD(day, @start_day_offset, t.cohort_start_date)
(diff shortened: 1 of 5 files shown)


azimov:
@gowthamrao I removed a number of columns from the results that appear to be unsused in the shiny modules and provide identical data to existing data. Would you please confirm that this change is ok?

inst/sql/sql_server/CohortRelationship.sql
```diff
@@ -24,79 +24,43 @@ SELECT t.cohort_definition_id cohort_id,
...
-			END) rec_cs_before_ts,
 	-- comparator cohort start date before target start date (offset) [How many subjects in comparator cohort start prior to first target cohort start]
 	COUNT_BIG(DISTINCT CASE 
 			WHEN c.cohort_start_date = DATEADD(day, @start_day_offset, t.cohort_start_date)
 				THEN c.subject_id
 			ELSE NULL
 			END) sub_cs_on_ts,
-	COUNT_BIG(DISTINCT CASE 
-			WHEN c.cohort_start_date = DATEADD(day, @start_day_offset, t.cohort_start_date)
-				THEN CONCAT(cast(c.subject_id AS VARCHAR(30)), '_', cast(c.cohort_start_date AS VARCHAR(30)))
```
azimov, line 40: @gowthamrao I'm not sure why this count would ever be different to the above

PR #1136, opened 2024-09-19
Use external concept counts

ablack3:
Moving this PR to the OHDSI organization so we can run all the tests in github actions.


R/RunDiagnostics.R
```diff
@@ -687,6 +692,37 @@ executeDiagnostics <- function(cohortDefinitionSet,
...
   }
+  
+  # Defines variables and checks version of external concept counts table -----
+  if (!useExternalConceptCountsTable) {
+    conceptCountsTableIsTemp <- TRUE
+    if (conceptCountsTable != "#concept_counts") {
+      conceptCountsTable <- "#concept_counts"
+    }
+  } else {
+    if (conceptCountsTable == "#concept_counts") {
```
azimov, line 703: This should probably be ```substr(conceptCountsTable, 1, 1) == "#"``` as otherwise `#my_temp_table` will render as `work_database_schema.#my_temp_table` which will crash on some systems

R/RunDiagnostics.R
```diff
@@ -687,6 +692,37 @@ executeDiagnostics <- function(cohortDefinitionSet,
...
+    vocabVersion <- dataSourceInfo$vocabularyVersion
+    vocabVersionExternalConceptCountsTable <- renderTranslateQuerySql(
+      connection = connection,
+      sql = "SELECT DISTINCT vocabulary_version FROM @work_database_schema.@concept_counts_table;",
+      work_database_schema = cohortDatabaseSchema,
+      concept_counts_table = conceptCountsTable,
+      snakeCaseToCamelCase = TRUE,
+      tempEmulationSchema = getOption("sqlRenderTempEmulationSchena")
+    )
+    if (!identical(vocabVersion, vocabVersionExternalConceptCountsTable[1,1])) {
```
azimov, line 718: great check!

tests/testthat/test-other.R
```diff
@@ -157,3 +157,44 @@ test_that("enforceMinCellValue works with vector of minimum values", {
 
   expect_equal(result$a, c(1, 2, 3, 4, 5))
 })
+
+test_that("Creating and checking externalConceptCounts table", {
+  if (dbms == "sqlite") {
+    # Creating externalConceptCounts
+    sql_lite_path <- file.path(test_path(), databaseFile)
+    connectionDetails <- createConnectionDetails(dbms= "sqlite", server = sql_lite_path)
```
azimov, line 165: We may want to build tests out for real databases but this can likely wait - I think adding a pre-computed counts table should speed up the other unit tests

azimov (approved):
Only minor changes suggested - add when ready.

A possible change if you want would be to make the CD concept counts table here identical to the fields needed in the achilles table - before I go and check @ablack3 @cebarboza, are they substantially different?

ablack3:
> A possible change if you want would be to make the CD concept counts table here identical to the fields needed in the achilles table - before I go and check @ablack3 @cebarboza, are they substantially different?

The achilles table we would like to use has "concept_id", "record_count", "descendant_record_count", and we would like to add "person_count" and "descendant_person_count" so we could use those column names.

https://github.com/OHDSI/Achilles/blob/247b86da872ceeb80b70d35901faa6795d3ab0d8/inst/sql/sql_server/analyses/create_result_concept_table.sql#L31

ablack3:
We are adding two parameters that depend on each other: `conceptCountsTable` (temp or not temp) and `useExternalConceptCountsTable` (T or F).

Let's list out the expected behavior:

|           | T                                                                        | F                                                    |
|-----------|--------------------------------------------------------------------------|------------------------------------------------------|
| temp      | error?                                                                   | created as temp table. replace if it already exists. |
| permanent | look for the table in the cohort schema. if it does not exist create it. | create temp table. replace if it already exists.     |

I'm wondering if we could do this with just one parameter, the cohortCountsTable. if it exists we use it and if not we recreate it.

PR #1172, opened 2025-04-22
Update SQL queries according to previous PR

Lui-M:
As a result of the previous comments in the PR, a new PR has been created.


Lui-M:
@azimov, here is a new PR created to address some of the issues from the previous one. Thank you.”

PR #1189, opened 2026-02-03
Refactor tests and project structure

azimov:


tests/testthat/setup.R
```diff
@@ -1,167 +1,230 @@
...
 
-if (dir.exists(Sys.getenv("DATABASECONNECTOR_JAR_FOLDER"))) {
-  jdbcDriverFolder <- Sys.getenv("DATABASECONNECTOR_JAR_FOLDER")
-} else {
-  jdbcDriverFolder <- "~/.jdbcDrivers"
-  dir.create(jdbcDriverFolder, showWarnings = FALSE)
-  DatabaseConnector::downloadJdbcDrivers("postgresql", pathToDriver = jdbcDriverFolder)
+# Check if we're running integration tests
+if (Sys.getenv("INTEGRATION_TESTS") == "TRUE") {
+  # ============================================================================
```
anthonysena, line 10: I'd suggest adding `testthat::skip_on_cran()` here when integration tests are enabled. This is simply a safeguard since I'd expect this value is not set on CRAN. Alternatively, you could add `testthat::skip_on_cran()` to the top of tests/testthat/test-integration-runner.R to avoid ever running on CRAN

anthonysena (commented):
My focus in this review is with the goal of getting CohortDiagnostics into CRAN. I left one comment specific to skipping integration tests and also wanted to note that CRAN checks may fail due to the lack of a `tests/testthat.R` entrypoint. Otherwise, the separation of tests into unit/integration looks good from my review.

